Issue #3, opened 2019-04-04
Create a basic auction contract with fixed price


cducrest:
Should run for two weeks and work with a whitelist.

PR #14, opened 2019-04-10
Add simple validator auction contract with whitelist and fixed price

cducrest:
closes https://github.com/trustlines-protocol/blockchain/issues/3
@berndbohmeier

### DIFF
--- a/contracts/tests/conftest.py
+++ b/contracts/tests/conftest.py
@@ -28,6 +28,12 @@ def deposit_amount():
 
 
 @pytest.fixture(scope="session")
+def number_of_auction_participants():
+    number_of_participants = 123
+    return number_of_participants
+
+
+@pytest.fixture(scope="session")
 def malicious_non_validator_address(accounts):
     return accounts[MALICIOUS_NON_VALIDATOR_INDEX]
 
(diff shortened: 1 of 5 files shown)


contracts/tests/test_validator_auction.py
```diff
@@ -0,0 +1,209 @@
...
+
+def test_auction_start_not_owner(validator_auction_contract, accounts):
+
+    with pytest.raises(eth_tester.exceptions.TransactionFailed):
+        validator_auction_contract.functions.startAuction().transact(
+            {"from": accounts[1]}
+        )
+
+
+def test_bidding_not_whiteliested(started_validator_auction, accounts):
```
schmir: type s/test_bidding_not_whiteliested/test_bidding_not_whitelisted/

contracts/tests/test_validator_auction.py
```diff
@@ -0,0 +1,209 @@
...
+
+    print(accounts[0])
+
+    with pytest.raises(eth_tester.exceptions.TransactionFailed):
+        started_validator_auction.functions.bid().transact(
+            {"from": accounts[0], "value": 100}
+        )
+
+
+def test_bidding_not_enough_ether(started_validator_auction, accounts):
```
schmir: How about calling this `test_bidding_bid_below_current_price`?

contracts/tests/conftest.py
```diff
@@ -153,3 +159,36 @@ def block_header_by_malicious_validator(malicious_validator_key):
...
+        chain.send_transaction({'from': account_0, 'to': extended_accounts[i], 'gas': 21000, 'value': 10000000})
+
+
+@pytest.fixture(scope="session")
+def extended_accounts(chain, number_of_auction_participants):
+    new_chain = chain
+    for i in range(100, 100 + number_of_auction_participants):
+        new_chain.add_account("0x0000000000000000000000000000000000000000000000000000000000000" + str(i))
+
+    return new_chain.get_accounts()
```
schmir: This should be changed to return only the possible participants. At the moment the rest of the code is littered with
```
for i in range(1, ...):
   do_something_with(extended_account[i])
 ```
It would be much nicer if we could iterate over the list directly instead of iterating over the indexes. It would become something like
```
for participant_account in extended_accounts:
  do_something_with(participant_account)
```
Also, I suggest using a better name. `participant_accounts` would be much clearer to me. 
And the last thing is,we generate exactly `number_of_auction_participants` accounts,but we will have more whitelisted accounts than participants. I think we should whitelist more accounts in the test setup.

cducrest:
> What is the purpose of 'closingBidValue'? It is the value that the last bidder did bid and it could be much higher than the value any previous bidder bid. (I could understand the purpose of closingBidValue if it was set to the currentPrice() active, when we got the bid filling the last available slot.

You are correct, having the closing price is useful while having the last bid is useless.

PR #16, opened 2019-04-12
Issue 5 deposit locker refactoring

schmir:
This contains a first version of the `DepositLocker` changes. I haven't testet it and I haven't adapted the tests.

### DIFF
--- a/contracts/tests/conftest.py
+++ b/contracts/tests/conftest.py
@@ -87,7 +87,6 @@ def non_initialised_deposit_locker_contract_session(deploy_contract):
 def initialised_deposit_and_slasher_contracts(validators, deploy_contract, web3):
     slasher_contract = deploy_contract("TestValidatorSlasher")
     locker_contract = deploy_contract("DepositLocker")
-
     """Initialises both the slasher and deposit contract, both initialisation are in the same fixture because we want
     a snapshot where both contracts are initialised and aware of the address of the other"""
 
@@ -99,9 +98,13 @@ def initialised_deposit_and_slasher_contracts(validators, deploy_contract, web3)
     # if this number is too high, tests are slowed down
 
     slasher_contract_address = slasher_contract.address
-
+    auction_contract_address = "0x0000000000000000000000000000000000000000"
     initialised_deposit_contract = initialize_deposit_locker(
-        locker_contract, release_number, slasher_contract_address, web3
+        locker_contract,
+        release_number,
+        slasher_contract_address,
+        auction_contract_address,
+        web3,
     )
 
     # initialise slasher contract
--- a/contracts/tests/conftest.py
+++ b/contracts/tests/conftest.py
@@ -34,7 +34,7 @@ def number_of_auction_participants():
 
 
 @pytest.fixture(scope="session")
-def auction_address(accounts):
+def fake_auction_address(accounts):
     return accounts[4]
 
 
@@ -90,7 +90,7 @@ def non_initialised_deposit_locker_contract_session(deploy_contract):
 
 @pytest.fixture(scope="session")
 def initialised_deposit_and_slasher_contracts(
-    validators, deploy_contract, auction_address, web3
+    validators, deploy_contract, fake_auction_address, web3
 ):
     slasher_contract = deploy_contract("TestValidatorSlasher")
     locker_contract = deploy_contract("DepositLocker")
@@ -105,7 +105,7 @@ def initialised_deposit_and_slasher_contracts(
     # if this number is too high, tests are slowed down
 
     slasher_contract_address = slasher_contract.address
-    auction_contract_address = auction_address
+    auction_contract_address = fake_auction_address
     initialised_deposit_contract = initialize_deposit_locker(
         locker_contract,
         release_number,
@@ -149,18 +149,18 @@ def deposit_locker_contract_with_deposits(
     validators,
     malicious_validator_address,
     deposit_amount,
-    auction_address,
+    fake_auction_address,
 ):
 
     deposit_contract = initialised_deposit_and_slasher_contracts.deposit_contract
 
     for validator in validators:
         deposit_contract.functions.registerDepositor(validator).transact(
-            {"from": auction_address}
+            {"from": fake_auction_address}
         )
 
     deposit_contract.functions.deposit(deposit_amount).transact(
-        {"from": auction_address, "value": deposit_amount * len(validators)}
+        {"from": fake_auction_address, "value": deposit_amount * len(validators)}
     )
 
     return deposit_contract
--- a/contracts/tests/conftest.py
+++ b/contracts/tests/conftest.py
@@ -156,11 +156,11 @@ def deposit_locker_contract_with_deposits(
     deposit_contract = initialised_deposit_and_slasher_contracts.deposit_contract
 
     for validator in validators:
-        deposit_contract.functions.registerParticipant(validator).transact(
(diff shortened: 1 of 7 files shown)


contracts/tests/conftest.py
```diff
@@ -33,6 +33,12 @@ def number_of_auction_participants():
     return number_of_participants
 
 
+@pytest.fixture(scope="session")
+def auction_address(accounts):
+    return accounts[4]
+    # Come said I should use 4. We should fix that index stuff here
```
cducrest: Then fix it before merging please.

contracts/tests/conftest.py
```diff
@@ -33,6 +33,12 @@ def number_of_auction_participants():
     return number_of_participants
 
 
+@pytest.fixture(scope="session")
+def auction_address(accounts):
```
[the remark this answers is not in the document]
schmir: I would rather go for `depositors_proxy` then, since this is what it's called in DepositLocker. ok?

contracts/tests/conftest.py
schmir: It's a different issue, I fail to see the need to fix it as part of this  PR.

contracts/tests/test_deposit_locker.py
```diff
@@ -213,5 +166,190 @@ def test_event_slash(
...
-    assert event["validator"] == malicious_validator_address
+    assert event["slashedDepositor"] == malicious_validator_address
     assert event["slashedValue"] == deposit_amount
+
+
+@attr.s(auto_attribs=True)
+class Env:
+    deposit_locker: Any
+    slasher: Any
+    proxy: Any
```
cducrest: rename to `depositors_proxy`.

I tried finding a better name for `depositors` and `depositors_proxy` as one could mix them up. I could not, so please keep the full name to avoid confusion.

contracts/tests/test_deposit_locker.py
```diff
@@ -213,5 +166,190 @@ def test_event_slash(
...
+    def withdraw(self, depositor):
+        return self.deposit_locker.functions.withdraw().transact({"from": depositor})
+
+    def can_withdraw(self, depositor):
+        return self.deposit_locker.functions.canWithdraw(depositor).call()
+
+
+@pytest.fixture()
+def testenv(deploy_contract, accounts, web3):
+    """return a initialized testenv instance"""
```
cducrest: return an initialized ...

contracts/tests/test_deposit_locker.py
```diff
@@ -213,5 +166,190 @@ def test_event_slash(
...
+    deposit_locker = deploy_contract("DepositLocker")
+    slasher = accounts[1]
+    proxy = accounts[2]
```
cducrest: shouldn't this use the `auction_address` fixture?

contracts/tests/test_deposit_locker.py
```diff
@@ -213,5 +166,190 @@ def test_event_slash(
...
+        )
+
+    def withdraw(self, depositor):
+        return self.deposit_locker.functions.withdraw().transact({"from": depositor})
+
+    def can_withdraw(self, depositor):
+        return self.deposit_locker.functions.canWithdraw(depositor).call()
+
+
+@pytest.fixture()
```
cducrest: This should be session scoped to avoid redeploying a locker contract for every test.

contracts/tests/test_deposit_locker.py
```diff
@@ -213,5 +166,190 @@ def test_event_slash(
...
+def testenv(deploy_contract, accounts, web3):
+    """return a initialized testenv instance"""
+    deposit_locker = deploy_contract("DepositLocker")
+    slasher = accounts[1]
+    proxy = accounts[2]
+    depositors = accounts[3:6]
+    other_accounts = accounts[6:]
+    deposit_locker.functions.init(web3.eth.blockNumber + 50, slasher, proxy).transact(
```
cducrest: You could also have tests showing how functions depending on values set in `init()` should fail if the contract is not initialised.

contracts/tests/test_deposit_locker.py
```diff
@@ -213,5 +166,190 @@ def test_event_slash(
...
+
+    def withdraw(self, depositor):
+        return self.deposit_locker.functions.withdraw().transact({"from": depositor})
+
+    def can_withdraw(self, depositor):
+        return self.deposit_locker.functions.canWithdraw(depositor).call()
+
+
+@pytest.fixture()
+def testenv(deploy_contract, accounts, web3):
```
cducrest: I guess this should be a `deposit_locker_testenv` ?

contracts/tests/conftest.py
cducrest: Then potentially remove the comment and create a github issue about cleaning this. The comment is not helping. (but really, I didn't believe it would be too much work to fix it, so creating an issue seems overkill.)

contracts/tests/test_deposit_locker.py
schmir: Other alternative would be delegate, trustee, custodian, agent.

schmir:
> I don't really understand why you created this whole env class, with the testenv fixture. It feels to me that pytest is already a test env and should allow to do what you are doing with fixtures.

I'm using a fixture. The Env class is just a way to group logically related values together. I could replace the single testenv fixture with 5 other fixtures that create the values that the Env class is made of and add another fixture to initialize the deposit locker contract. I would argue that what we have in this PR is much simpler.

> Maybe it is more convenient to use, but then we should probably do that for every contract, and not have a mix of raw pytest fixtures / testenv classes.

We already have that mix. As you can see in the following code, a combined object is being created and it's even ripped apart through with more fixtures, that extract a single value.

https://github.com/trustlines-protocol/blockchain/blob/427d2ca9683826cf3ae109ae62c97d9c9ef2e37e/contracts/tests/conftest.py#L86-L133

contracts/tests/test_deposit_locker.py
schmir: no, the fixture lives in a file called test_deposit_locker. That is enough.

contracts/tests/test_deposit_locker.py
schmir: 3fcc727a5783c3ce01c0adecce826424cc5cdda0

contracts/tests/test_deposit_locker.py
schmir: 3955fc075da4f7640d46eb84a11b65602cec0f09

contracts/tests/conftest.py
schmir: a46e2749a89b62a3bdf8a289f88d1faec059

contracts/tests/conftest.py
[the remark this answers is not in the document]
schmir: 3789c5168781dd9d772aa6fa3be28168fc7e592b

contracts/tests/test_deposit_locker.py
schmir: proxy is being used in the context of the Env class. I fail to see any confusion arising from that.

contracts/tests/test_deposit_locker.py
schmir: e458e701a90b59e31ef1ff3663751056f063c1e5

schmir: though I do not think they provide much value. One could safely remove the isInitialized modifier. If you tried to call any of the functions they would fail for different reasons.

PR #19, opened 2019-04-15
Feature/withdraw failed auction

cducrest:
closes https://github.com/trustlines-protocol/blockchain/issues/17

Based on https://github.com/trustlines-protocol/blockchain/pull/18

### DIFF
--- a/contracts/contracts/ValidatorAuction.sol
+++ b/contracts/contracts/ValidatorAuction.sol
@@ -92,12 +92,15 @@ contract ValidatorAuction is Ownable {
         return 1;
     }
 
-    function withdrawOverbid() public stateIs(AuctionStates.Ended) {
+    function withdraw() public {
+        require(auctionState == AuctionStates.Ended || auctionState == AuctionStates.Failed, "You cannot withdraw before the auction is ended or it failed.");
         require(bids[msg.sender] > closingPrice, "The sender has nothing to withdraw.");
-        uint overbid = bids[msg.sender] - closingPrice;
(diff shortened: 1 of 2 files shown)


contracts/contracts/ValidatorAuction.sol
```diff
@@ -92,6 +92,15 @@ contract ValidatorAuction is Ownable {
         return 1;
     }
 
+    function withdraw() public {
+        require(auctionState == AuctionStates.Ended || auctionState == AuctionStates.Failed, "You cannot withdraw before the auction is ended or it failed.");
+        require(bids[msg.sender] > closingPrice, "The sender has nothing to withdraw.");
+        uint overbid = bids[msg.sender] - closingPrice;
+        assert(overbid <= bids[msg.sender]);
```
schmir: I do think this assert is confusing because it's true anyway (if closingPrice is not negative) and it's not 'strong enough' in the sense that with `overbid=bids[msg.sender]` that assert would not trigger, even though the whole bid would be withdrawn. Can we just remove it? Or did I miss the point somehow?

cducrest: The idea about asserts is that we could just remove them. It is just here as a safety to make sure we do not underflow, even though I cannot see any way of it happening, it might prevent an exploit in some weird way I cannot foresee.

We want the whole bid to be withdrawn if `overbid=bids[msg.sender]`, because it means the auction failed and participants should be able to withdraw everything they bid.

Issue #20, opened 2019-04-16
Add event for whitelisted addresses of auction


cducrest:
Need to emit an event when addresses are whitelisted.
Potentially one event per addresses or one event for the a whole list of addresses.

PR #22, opened 2019-04-17
Add event for adding addresses to whitelist

cducrest:
closes https://github.com/trustlines-protocol/blockchain/issues/20

### DIFF
--- a/contracts/tests/conftest.py
+++ b/contracts/tests/conftest.py
@@ -1,6 +1,7 @@
-import pytest
-
 from collections import namedtuple
+
+import pytest
+import eth_tester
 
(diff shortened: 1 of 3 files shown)


contracts/tests/conftest.py
```diff
@@ -10,6 +11,11 @@
 
 from .data_generation import make_block_header
 
+# increase eth_tester's GAS_LIMIT
+# Otherwise we can't deploy our contract
```
cducrest: We can deploy it, we just can't whitelist 123 validators at the same time.

cducrest:
Looks good to me

PR #25, opened 2019-04-18
Make price function configurable

berndbohmeier:
this should make the auction better testable

### DIFF
--- a/contracts/contracts/ValidatorAuction.sol
+++ b/contracts/contracts/ValidatorAuction.sol
@@ -13,7 +13,8 @@ contract ValidatorAuction is Ownable {
 
     uint public startTime;
     uint public closeTime;
-    uint public constant AUCTION_DURATION = 14 days;
+    uint public auctionDurationInDays;
+    uint public startPrice;
     uint public constant NUMBER_OF_PARTICIPANTS = 123;
 
     event BidSubmitted(address bidder, uint bidValue, uint slotPrice, uint timestamp);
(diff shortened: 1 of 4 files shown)


contracts/contracts/ValidatorAuction.sol
```diff
@@ -13,7 +13,8 @@ contract ValidatorAuction is Ownable {
 
     uint public startTime;
     uint public closeTime;
-    uint public constant AUCTION_DURATION = 14 days;
+    uint public auctionDurationInDays;
+    uint public startPrice;
     uint public constant NUMBER_OF_PARTICIPANTS = 123;
```
cducrest: Might as well parametrise the number of participants

contracts/tests/conftest.py
```diff
@@ -25,6 +25,9 @@
 MALICIOUS_VALIDATOR_INDEX = HONEST_VALIDATOR_COUNT
 MALICIOUS_NON_VALIDATOR_INDEX = MALICIOUS_VALIDATOR_INDEX + 1
 
+AUCTION_DURATION_IN_DAYS = 14
+START_PRICE_IN_ETH = 10000
```
cducrest: AUCTION_START_PRICE_IN_ETH

andfletcher:
https://github.com/trustlines-network/project/issues/319

contracts/contracts/ValidatorAuction.sol
berndbohmeier: yeah, but I thought we dont need it really. Do you think I should do it?

contracts/tests/conftest.py
berndbohmeier: done

contracts/contracts/ValidatorAuction.sol
berndbohmeier: okay

contracts/tests/test_validator_auction.py
```diff
@@ -346,7 +360,7 @@ def test_event_whitelist(no_whitelist_validator_auction_contract, whitelist, web
 
 
 ETH_IN_WEI = 1e18
-START_PRICE = 10000 * ETH_IN_WEI
+AUCTION_START_PRICE = 10000 * ETH_IN_WEI
```
cducrest: Should this not be at the beginning of the file?
Also you have duplication of `AUCTION_START_PRICE`

berndbohmeier: Fixed

Issue #5, opened 2019-04-08
Send auction ether to locking contract


cducrest:
The validator contract should be able to send `123 * lowest_bid` ether to the locking contract along with the winner addresses once the auction is over.

schmir:
The current plan is:
- remove the deposit function from the `DepositLocker` contract
- add a `registerParticipant` function, which registers a participant. This will be called from the auction contract, when a successful bid came in with the participants address.
- make sure the above `registerParticipant` function can only be called from the auction contract , therefore we need to store the address of the auction contract inside the deposit locker contract.
- add a function `depositBids` to the auction contract that transfers `NUMBER_OF_PARTICIPANTS * slot_price` eth to the deposit contract
- add a function to the depositContract that takes those bids and distributes it to the registered participants

cducrest:
> add a function to the depositContract that takes those bids and distributes it to the registered participants. This should only be callable from the auction contract.

I don't believe you need to distribute the bids to the registered participants in the deposit contract.
You can more easily keep track of who is able to withdraw deposit by initiating this with the successful bidders and removing addresses as they withdraw their deposits or get slashed.

I will add that the auction contract needs to know the address of the locker contract to know what contract to call.

schmir:
Thanks @cducrest. I'll start with the changes to the DepositLocker contract then.

schmir:
The deposit locker changes are in this PR https://github.com/trustlines-protocol/blockchain/pull/16

berndbohmeier:
Done

PR #26, opened 2019-04-18
Feature/deposit in locker

schmir:
This is the last part to fully implement https://github.com/trustlines-protocol/blockchain/issues/5

There's some code duplication when deploying auction contracts in conftest.py. I'd like to handle that later.

### DIFF
--- a/contracts/tests/conftest.py
+++ b/contracts/tests/conftest.py
@@ -183,7 +183,7 @@ def block_header_by_malicious_non_validator(malicious_non_validator_key):
 
 
 @pytest.fixture(scope="session")
-def validator_auction_contract(deploy_contract, whitelist, web3, default_account):
+def validator_auction_contract(deploy_contract, whitelist, web3):
     deposit_locker = deploy_contract("DepositLocker")
     contract = deploy_contract(
         "TestValidatorAuctionFixedPrice", constructor_args=(deposit_locker.address,)
@@ -192,7 +192,7 @@ def validator_auction_contract(deploy_contract, whitelist, web3, default_account
         web3.eth.blockNumber + 50,
         "0x0000000000000000000000000000000000000000",
         contract.address,
-    ).transact({"from": default_account})
+    ).transact()
 
     add_whitelist_to_validator_auction_contract(contract, whitelist)
 
@@ -200,9 +200,7 @@ def validator_auction_contract(deploy_contract, whitelist, web3, default_account
 
 
 @pytest.fixture(scope="session")
-def real_price_validator_auction_contract(
-    deploy_contract, whitelist, default_account, web3
-):
+def real_price_validator_auction_contract(deploy_contract, whitelist, web3):
     deposit_locker = deploy_contract("DepositLocker")
 
     contract = deploy_contract(
@@ -212,7 +210,7 @@ def real_price_validator_auction_contract(
         web3.eth.blockNumber + 50,
         "0x0000000000000000000000000000000000000000",
         contract.address,
-    ).transact({"from": default_account})
+    ).transact()
 
     add_whitelist_to_validator_auction_contract(contract, whitelist)
 
@@ -220,7 +218,7 @@ def real_price_validator_auction_contract(
 
 
 @pytest.fixture(scope="session")
-def no_whitelist_validator_auction_contract(deploy_contract, web3, default_account):
+def no_whitelist_validator_auction_contract(deploy_contract, web3):
     deposit_locker = deploy_contract("DepositLocker")
     contract = deploy_contract(
         "TestValidatorAuctionFixedPrice", constructor_args=(deposit_locker.address,)
@@ -229,14 +227,14 @@ def no_whitelist_validator_auction_contract(deploy_contract, web3, default_accou
         web3.eth.blockNumber + 50,
         "0x0000000000000000000000000000000000000000",
         contract.address,
-    ).transact({"from": default_account})
+    ).transact()
 
     return contract
 
 
 @pytest.fixture(scope="session")
 def almost_filled_validator_auction(
-    deploy_contract, whitelist, number_of_auction_participants, web3, default_account
+    deploy_contract, whitelist, number_of_auction_participants, web3
 ):
     """Validator auction contract missing one bid to reach the maximum amount of bidders
     account[1] has not bid and can be used to test the behaviour of sending the last bid"""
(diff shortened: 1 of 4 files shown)


contracts/contracts/TestValidatorAuctionFixedPrice.sol
```diff
@@ -9,6 +9,8 @@ import "./ValidatorAuction.sol";
 
 
 contract TestValidatorAuctionFixedPrice is ValidatorAuction {
+    constructor(DepositLocker _depositLocker) ValidatorAuction(_depositLocker) {
```
cducrest: I'd rather we do it the way it is done in Bernd's PR: https://github.com/trustlines-protocol/blockchain/pull/25/files#diff-ec429f556412759d5ab5e44787fb4a82R11
Mostly for consistency but also because it is the recommanded way

contracts/tests/conftest.py
```diff
@@ -183,36 +183,74 @@ def block_header_by_malicious_non_validator(malicious_non_validator_key):
...
+    deposit_locker = deploy_contract("DepositLocker")
+
+    contract = deploy_contract(
+        "ValidatorAuction", constructor_args=(deposit_locker.address,)
+    )
+    deposit_locker.functions.init(
+        web3.eth.blockNumber + 50,
+        "0x0000000000000000000000000000000000000000",
+        contract.address,
+    ).transact({"from": default_account})
```
cducrest: The default_account is used per default, no need to specify it

contracts/tests/conftest.py
```diff
@@ -183,36 +183,74 @@ def block_header_by_malicious_non_validator(malicious_non_validator_key):
...
+    deploy_contract, whitelist, default_account, web3
+):
+    deposit_locker = deploy_contract("DepositLocker")
+
+    contract = deploy_contract(
+        "ValidatorAuction", constructor_args=(deposit_locker.address,)
+    )
+    deposit_locker.functions.init(
+        web3.eth.blockNumber + 50,
+        "0x0000000000000000000000000000000000000000",
```
cducrest: Please give a name to the arguments so we don't have to go to the contract to see what they are.

contracts/tests/conftest.py
```diff
@@ -183,36 +183,74 @@ def block_header_by_malicious_non_validator(malicious_non_validator_key):
...
+def real_price_validator_auction_contract(
+    deploy_contract, whitelist, default_account, web3
+):
+    deposit_locker = deploy_contract("DepositLocker")
+
+    contract = deploy_contract(
+        "ValidatorAuction", constructor_args=(deposit_locker.address,)
+    )
+    deposit_locker.functions.init(
+        web3.eth.blockNumber + 50,
```
cducrest: Use `RELEASE_BLOCK_NUMBER_OFFSET` constant instead of 50. If we have too much transactions in conftest, we might have to increase this number in the future.

contracts/tests/test_validator_auction.py
```diff
@@ -158,12 +159,34 @@ def test_bidding_auction_ended(started_validator_auction, accounts, chain):
...
         {"from": accounts[1], "value": 100}
     )
 
+    assert_auction_state(almost_filled_validator_auction, AuctionStates.DepositPending)
+
+    deposit_locker = almost_filled_validator_auction.functions.depositLocker().call()
+    assert web3.eth.getBalance(deposit_locker) == 0
+
+    # we could test the rest of this in another function, but it's slow anyway
+    # and I'm lazy
```
cducrest: :(

contracts/contracts/TestValidatorAuctionFixedPrice.sol
schmir: Can you please write that down for me? I don't know how this is supposed to be done. Bernds PR only uses constant values. I don't know how to pass that argument down to the base constructor.

contracts/tests/conftest.py
schmir: 760af1bab84ca6aa80b60bc59c4761e6e594f822

contracts/tests/conftest.py
schmir: 534398b935d5fc87d8f2bbccddf227d7a5a7c5a3

contracts/tests/conftest.py
schmir: dd4e6e44d2ebfa6960d32f82744dddcdfc8485

contracts/contracts/TestValidatorAuctionFixedPrice.sol
cducrest: You're right, it actually does not work, sorry.

contracts/tests/test_validator_auction.py
schmir: The test goes through the process of ending an auction. I fail to see the benefits of creating two tests out of this, other than satisfying the 'a test should only test one thing rule', which I'm willing to break. 

If you think this must be changed, please a) say so (a sad smiley is not enough) and b) state the reasons why.

cducrest:
It seemed to me that the reason you didn't split the test in two was because you were lazy, which I do not believe is a good reason. I thought you agreed on the fact that this test should be split into two. 

I believed I needed to change your PR anyway to remove this useless comment, and since I had the code open, I might as well split the test in two.

However, I agree that splitting it into two introduces the problems that you described in your comment. I thought about them and thought that I potentially should add more unit tests to cover them but did not, since I do not believe the errors prevented by these tests will go unnoticed.

I then attempted to push the commit to your PR, as proposed by github, but it failed and created a new branch here, which is the reason why I opened a new PR (to potentially gain some time if you just wanted to merge it). Seeing it had conflicts, I decided to rebase it, in the same intent of allowing to merge it quickly.

I still don't see why it is wrong, but it seems we do not care about the same things, so merge your PR if you are fine with it. I guess someone will remove the useless comment the next time he reads them.

schmir-at-bb:
> It seemed to me that the reason you didn't split the test in two was because you were lazy, which I do not believe is a good reason. 

The comment actually stated two reasons for not doing it.

> I thought you agreed on the fact that this test should be split into two.

No, the opposite is true. I was pondering if I should split this test and decided against it, because I didn't see any clear benefit in doing so. I left the comment in there to actually not be bothered at all about splitting it.

I told you in the daily that I consider it a waste of time, yet the change has to be done.

To me this is just a drain of mental energy for some minor technical improvements (if at all).

PR #38, opened 2019-04-23
Schmir forks feature/deposit in locker

cducrest:
Apparently Ralf has to allow me to push to his PR otherwise I cannot do it  (https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) so I opened a new PR with the quick fix. Even though I cannot see this option suggested by gh.

Just merge this Ralf if you are fine, I resolved the conflicts.

### DIFF
--- a/contracts/contracts/TestValidatorAuctionFixedPrice.sol
+++ b/contracts/contracts/TestValidatorAuctionFixedPrice.sol
@@ -9,6 +9,8 @@ import "./ValidatorAuction.sol";
 
 
 contract TestValidatorAuctionFixedPrice is ValidatorAuction {
+    constructor(DepositLocker _depositLocker) ValidatorAuction(_depositLocker) {
+    }
 
     function currentPrice() public view returns (uint) {
         return 100;
--- a/contracts/contracts/ValidatorAuction.sol
+++ b/contracts/contracts/ValidatorAuction.sol
@@ -26,6 +26,7 @@ contract ValidatorAuction is Ownable {
     enum AuctionStates{
         Deployed,
         Started,
+        DepositPending, /* all slots sold, someone needs to call depositBids */
         Ended,
         Failed
     }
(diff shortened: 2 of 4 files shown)


schmir:
Please do not rebase my changes. It's hard to see what's going on that way and we lose the PR history. You should have opened a PR to my branch. If that's too much work, you could also let me merge my PR and open a new one that fixes the problem you seem to have with my code.

Unfortunately you still fail to explain the reasoning why we need to split that test into two - even though I asked for an explanation in the other PR. If your only argument is "unit tests should only test one thing", we could say it's an integration test and be done with it. Running `pytest -k test_enough_bidders_ends_auction` takes 50 seconds on my machine for running that single test, that's certainly too much for a unit test.

I would even argue that what we have here is in some way worse than what we had before. If placing the bid on `almost_filled_validator_auction` does not put the auction into the state `DepositPending`, then 
`test_enough_bidders_ends_auction` will fail. 
But since you removed the assert about the expected state, `test_send_bids_to_locker` will continue and try to call `depositBids`, which will then fail because the auction is in the wrong state. But, we should have better failed early in `test_send_bids_to_locker`.  Please add the `assert_auction_state(almost_filled_validator_auction, AuctionStates.DepositPending)` again.

Also, you introduced some probelms during rebasing. The last commit removes a duplicate  AuctionStates definition, which was not part of my branch. I think I fixed this within my merge of the master branch.
circle ci would have complained about that error when running flake8. The ci history also get's lost when rebasing. So,again please don't rebase.

I suggest choosing a different name for `test_enough_bidders_ends_auction`.  The auction state is not `Ended` after the bid that fills the auction, it's `DepositPending`.

In order to not waste even more time, I suggest merging my PR as is, then you could open a new PR for the problems that you care about (please provide some reasoning and take my suggestions into account).

If you don't want to have my PR merged as is, please at least provide some reasoning in the original PR.

Please close this one or base it on my non-rebased changes.

berndbohmeier:
The rabseing was not good, I agree. But I also agree that we should split up the tests. It is not much work, so I dont know why you are against it. 
We will close this PR and reopen another unrebased branch with the changes.

PR #41, opened 2019-04-24
Extend deposit locker slashing to burn deposit

### LINKED ISSUE #36: Burn eth on slashing validators

weilbith:
Closes #36 

Notes:
- Please mark that the zero address is the one to burn to as well as the mining address.
- Since it should be never possible to have not enough Ether on the contract to burn, I haven't add any further `require` checks. This wouldn't be even possible to test with the current setup I think.

### DIFF
--- a/contracts/tests/test_deposit_locker.py
+++ b/contracts/tests/test_deposit_locker.py
@@ -358,14 +358,18 @@ def test_slash_from_non_slasher_throws(testenv):
         )
 
 
-def test_slash_after_deposit_works(
+def test_slash_after_deposit_can_not_withdraw(testenv_deposited):
+    depositor = testenv_deposited.depositors[0]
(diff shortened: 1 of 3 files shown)


contracts/tests/test_deposit_locker.py
```diff
@@ -353,11 +358,37 @@ def test_slash_from_non_slasher_throws(testenv):
         )
 
 
-def test_slash_after_deposit_works(testenv_deposited):
+def test_slash_after_deposit_works(
```
cducrest: Can you split these two tests: 
- one checks that the depositor cannot withdraw
- the second checks that the eth was burned.

weilbith: I've done so. :+1:

Issue #39, opened 2019-04-24
Prevent auction start if deposit not init


cducrest:
We should prevent starting the validator auction if the deposit contract has not been initialised.

PR #43, opened 2019-04-24
Feature/prevent auction start deposit not init

cducrest:
closes https://github.com/trustlines-protocol/blockchain/issues/39


contracts/contracts/DepositLocker.sol
```diff
@@ -45,7 +45,7 @@ contract DepositLocker is DepositLockerInterface, Ownable {
     event Slash(address slashedDepositor, uint slashedValue);
 
     modifier isInitialised() {
-        require(initialised, "The contract was not initialised.");
+        require(initialized, "The contract was not initialised.");
```
berndbohmeier: Please then also change the message

PR #49, opened 2019-04-30
cli to deploy auction

berndbohmeier:


auction-deploy/auction_deploy/core.py
```diff
@@ -0,0 +1,132 @@
...
+
+class AuctionOptions(NamedTuple):
+    start_price: int
+    auction_duration: int
+    number_of_participants: int
+    release_block_number: int
+
+
+DeployedAuctionContracts = namedtuple("DeployedContracts", "locker slasher auction")
+
```
schmir-at-bb: Please change this to use a subclass of NamedTuple like it is done a few lines above for the AuctionOptions class (in order to keep it consistent in a single file).

auction-deploy/auction_deploy/core.py
```diff
@@ -0,0 +1,132 @@
...
+
+def decrypt_private_key(keystore: str, password: str) -> bytes:
+    return extract_key_from_keyfile(keystore, password.encode("utf-8"))
+
+
+def increase_transaction_options_nonce(transaction_options: Dict) -> None:
+    if "nonce" in transaction_options:
+        transaction_options["nonce"] = transaction_options["nonce"] + 1
+
```
schmir-at-bb: It might be better to throw an exception if "nonce" isn't contained in `transactions_options`. I'm not sure however how this function is being used,so it might also be done on purpose. In that case,please add a docstring.

auction-deploy/auction_deploy/core.py
```diff
@@ -0,0 +1,132 @@
...
+    number_of_participants: int
+    release_block_number: int
+
+
+DeployedAuctionContracts = namedtuple("DeployedContracts", "locker slasher auction")
+
+
+def load_contracts_json() -> Dict:
+    resource_package = __name__
+    stream = pkg_resources.resource_stream(resource_package, "contracts.json")
```
schmir-at-bb: from reading the setuptools documentation,  I think you should be able to use `pkg_resources.resource_string`. Please try to use that.

auction-deploy/auction_deploy/core.py
berndbohmeier: done

auction-deploy/auction_deploy/core.py
berndbohmeier: Added a doc string

auction-deploy/auction_deploy/core.py
berndbohmeier: done

Issue #34, opened 2019-04-23
Command to print the configuration of an auction


berndbohmeier:
It should print the settings of an auction so that we can verify that everything is correct.

PR #51, opened 2019-05-02
Feature/print auction settings

cducrest:
closes https://github.com/trustlines-protocol/blockchain/issues/34

### DIFF
--- a/auction-deploy/auction_deploy/cli.py
+++ b/auction-deploy/auction_deploy/cli.py
@@ -6,9 +6,11 @@ from auction_deploy.core import (
     initialize_auction_contracts,
     decrypt_private_key,
     AuctionOptions,
+    load_contracts_json,
 )
 from web3 import Web3, EthereumTesterProvider
 
+test_json_rpc = Web3(EthereumTesterProvider())
 
 jsonrpc_option = click.option(
     "--jsonrpc",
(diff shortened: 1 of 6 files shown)


auction-deploy/auction_deploy/cli.py
```diff
@@ -141,3 +153,81 @@ def deploy(
...
+    start_time = contracts.auction.functions.startTime().call()
+    close_time = contracts.auction.functions.closeTime().call()
+    closing_price = contracts.auction.functions.closingPrice().call()
+    if auction_state == AuctionState.Started:
+        current_price_in_eth = (
+            contracts.auction.functions.currentPrice().call() / ETH_IN_WEI
+        )
+
+    click.echo(
+        "The auction duration is:                " + str(duration_in_days) + " days"
```
berndbohmeier: I think in general we should better use string interpolation, so either `"this is a {}".format(variable)"` or `f"this is a {variable}"

contracts/contracts/ValidatorAuction.sol
```diff
@@ -118,8 +118,8 @@ contract ValidatorAuction is Ownable {
         }
     }
 
-    function currentPrice() public view returns (uint) {
-        require(now >= startTime, "The current price can only be calculated after the auction started.");
+    function currentPrice() public view stateIs(AuctionState.Started) returns (uint) {
```
berndbohmeier: If you fix something like this, we should add a test I think

PR #53, opened 2019-05-03
add --auto-nonce switch

schmir-at-bb:
This will make the program ask for the next nonce to be used via JSONRPC.

We need to provide the nonce to use infura.

### DIFF
--- a/auction-deploy/tests/test_cli.py
+++ b/auction-deploy/tests/test_cli.py
@@ -66,6 +66,20 @@ def test_cli_start_auction(runner, deployed_auction_address):
     assert result.exit_code == 0
 
 
+def test_cli_start_auction_with_auto_nonce(
+    runner, deployed_auction_address, keystores, key_password
+):
+    """test the auto-nonce option. we only do this for the start-auction"""
+
+    result = runner.invoke(
+        main,
+        args=f"start-auction --jsonrpc test --keystore {keystores[0]} --auction-address {deployed_auction_address}",
+        input=key_password,
+    )
+    print(result.output)
+    assert result.exit_code == 0
+
+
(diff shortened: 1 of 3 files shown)


schmir-at-bb:
Added some tests. Please have a look at it again.

auction-deploy/tests/test_cli.py
```diff
@@ -66,6 +66,20 @@ def test_cli_start_auction(runner, deployed_auction_address):
...
 
 
+def test_cli_start_auction_with_auto_nonce(
+    runner, deployed_auction_address, keystores, key_password
+):
+    """test the auto-nonce option. we only do this for the start-auction"""
+
+    result = runner.invoke(
+        main,
+        args=f"start-auction --jsonrpc test --keystore {keystores[0]} --auction-address {deployed_auction_address}",
```
cducrest: Does it make sense that there is no `auto-nonce` in the args here ?

schmir-at-bb: Right, `--auto-nonce` should have been there.

The print statement might provide useful information in case of an error. If the test succeeds, you won't see the output anyway.

I will open another PR as I've already merged this one.

schmir-at-bb: https://github.com/trustlines-protocol/blockchain/pull/59

schmir-at-bb:
> The auto-nonce might also be misinterpreted by someone not using a local private key, he will want to nonce to be automatically filled by the web3 provider but will get an error if he uses --auto-nonce.

I don't know how to get the default address of the unlocked account if there is any (and didn't care enough to look it up). If we know how to do that, we can add the code into the `get_nonce` function.

Though I do think it's not worth it. We will use a local private key anyway.

PR #56, opened 2019-05-03
Add validation for address input of cli

cducrest:
We wanted to merge the other PR to facilitate cooperation, so I opened this new PR.

### DIFF
--- a/auction-deploy/auction_deploy/cli.py
+++ b/auction-deploy/auction_deploy/cli.py
@@ -4,6 +4,7 @@ from enum import Enum
 import click
 from web3 import Web3, EthereumTesterProvider
 from deploy_tools.deploy import send_function_call_transaction
+from eth_utils import is_address, to_canonical_address
 
 from auction_deploy.core import (
     deploy_auction_contracts,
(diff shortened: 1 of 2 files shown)


auction-deploy/auction_deploy/cli.py
```diff
@@ -15,6 +16,16 @@
 )
 
 
+def validate_address(ctx, param, value):
+    """This function must be at the top of click commands using it"""
+    try:
+        return to_checksum_address(value)
```
berndbohmeier: I think we should better enforce a checksummed address for safety, so

auction-deploy/tests/test_cli.py
```diff
@@ -94,3 +96,33 @@ def test_cli_auction_status(runner, deployed_auction_address):
     )
 
     assert result.exit_code == 0
+
+
+def test_cli_address_not_prefixed_hex(runner, deployed_auction_address):
+
+    hex_prefix_size = len("0x")
+
+    address = deployed_auction_address[hex_prefix_size:]
```
berndbohmeier: Why do you remove the 0x?

cducrest: To see that it still works without, I don't care if we do not have it. I would rather be open when validating the address rather than strict.
If we enforce checksummed address then I'll remove this test.

auction-deploy/auction_deploy/cli.py
```diff
@@ -15,6 +16,16 @@
 )
 
 
+def validate_address(ctx, param, value):
+    """This function must be at the top of click commands using it"""
+    if is_address(value):
+        return to_checksum_address(value)
```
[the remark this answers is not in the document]
cducrest: I changed it to a canonical address.

PR #57, opened 2019-05-05
Add cli tool for whitelisting

berndbohmeier:
Add cli tool to read in a csv file(for now just list of addresses) with
addresses and add the diff of this list with the already whitelisted
addresses to the contract whitelist. The number of addresses within one
transaction is configurable

Fix some typings errors

### DIFF
--- a/auction-deploy/auction_deploy/cli.py
+++ b/auction-deploy/auction_deploy/cli.py
@@ -3,7 +3,6 @@ from enum import Enum
 import click
 from web3 import Web3, EthereumTesterProvider, Account
 from deploy_tools.deploy import send_function_call_transaction
-from eth_utils import is_address, to_checksum_address
 
 from auction_deploy.core import (
     deploy_auction_contracts,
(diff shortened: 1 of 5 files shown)


auction-deploy/auction_deploy/core.py
```diff
@@ -170,3 +172,66 @@ def get_deployed_auction_contracts(
...
+
+
+def read_whitelist(file_path: str):
+    with open(file_path) as f:
+        reader = csv.reader(f)
+        addresses = []
+        for line in reader:
+            address = line[0]
+            if not is_checksum_address(address):
+                raise ValueError("Invalid address in whitelist file")
```
cducrest: We could probably use the same validation strategy as in `validate_address` now that it has been merged (https://github.com/trustlines-protocol/blockchain/blob/master/auction-deploy/auction_deploy/cli.py#L19)

auction-deploy/auction_deploy/core.py
```diff
@@ -170,3 +172,66 @@ def get_deployed_auction_contracts(
     )
 
     return deployed_auction_contracts
+
+
+def whitelist_addresses(
+    auction_contract: Contract,
+    whitelist: Sequence[str],
+    *,
+    batch_size=100,
```
cducrest: I don't like having a default here since you already have one in the cli

auction-deploy/auction_deploy/cli.py
```diff
@@ -255,6 +251,62 @@ def print_auction_status(auction_address, jsonrpc):
...
+    help="Number of addresses to be whitelisted within one transaction",
+    type=click.IntRange(min=1),
+    show_default=True,
+    default=100,
+)
+@keystore_option
+@gas_option
+@gas_price_option
+@nonce_option
+@jsonrpc_option
```
cducrest: add `auto-nonce` option, now that it is merge with master

auction-deploy/auction_deploy/core.py
berndbohmeier: yes

auction-deploy/auction_deploy/cli.py
berndbohmeier: ok

auction-deploy/auction_deploy/core.py
berndbohmeier: ok

PR #58, opened 2019-05-06
Feature/issue 32 close auction

schmir-at-bb:
This implements the close-auction command. It has some useful fixtures that can be used to ensure the auction is in a certain state.

### DIFF
--- a/auction-deploy/auction_deploy/cli.py
+++ b/auction-deploy/auction_deploy/cli.py
@@ -14,8 +14,12 @@ from auction_deploy.core import (
     get_deployed_auction_contracts,
 )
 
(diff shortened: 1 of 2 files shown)


auction-deploy/auction_deploy/cli.py
```diff
@@ -202,6 +206,50 @@ def start_auction(
     )
 
 
+@main.command(short_help="Close the auction at corresponding address.")
+@click.option(
+    "--auction-address",
+    help='The address of the auction contract to be closed, "0x" prefixed string',
```
cducrest: add address validation now that it works with master

PR #66, opened 2019-05-07
Feature/checkwhitelist

berndbohmeier:


auction-deploy/auction_deploy/cli.py
```diff
@@ -78,6 +79,22 @@ class AuctionState(Enum):
...
+auction_address_option = click.option(
+    "--auction-address",
+    help='The address of the auction contract, "0x" prefixed string',
+    type=str,
+    required=True,
+    callback=validate_address,
+    metavar="ADDRESS",
+)
+whitelist_file_option = click.option(
+    "--file",
```
cducrest: Please use: `"--file", "whitelist_file"` for readability

auction-deploy/tests/test_cli.py
```diff
@@ -36,6 +36,19 @@ def deployed_auction_address(runner):
     raise ValueError(f"Could not find auction address in output: {lines}")
 
 
+@pytest.fixture()
+def whitelisted_auction_address(runner, deployed_auction_address, whitelist_file):
+    """deploy an auction with all addresses whitelisted and return it's address"""
```
cducrest: its*

cducrest: Also the comment lies, since this fixture does not deploy an auction but reuse the already deployed auction.
Since we have no clean_chain fixture, this is important.
(maybe we should start having a clean_chain fixture)

auction-deploy/auction_deploy/cli.py
berndbohmeier: I'll fix this in the simplify auction command PR

auction-deploy/tests/test_cli.py
berndbohmeier: as those fixtures are not bound to sessions it should actually always deploy a new contract for each test.

berndbohmeier: and we have a clean chain fixture? It is in the deploy tools plugin

berndbohmeier: And I copied the it's so will also change it there.

cducrest: The clean chain fixture is not cleaning the right chain, but yeah as long as we have no session scoped fixtures it is fine

berndbohmeier: ah yes you are right.

Issue #64, opened 2019-05-07
Change block number to timestamp in locker contract


cducrest:
Use timestamp instead of block nr
related: https://github.com/trustlines-network/project/issues/445

PR #68, opened 2019-05-08
Issue 64 release timestamp instead of blocknumber

schmir-at-bb:
This implements https://github.com/trustlines-protocol/blockchain/issues/64

There's no date parsing yet, I will open another PR for that.

### DIFF
--- a/contracts/tests/test_deposit_locker.py
+++ b/contracts/tests/test_deposit_locker.py
@@ -10,55 +10,53 @@ from web3.datastructures import AttributeDict
 
 @pytest.fixture()
 def deposit_contract_on_longer_chain(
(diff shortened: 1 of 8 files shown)


contracts/tests/test_deposit_locker.py
```diff
@@ -10,55 +10,53 @@
 
 @pytest.fixture()
 def deposit_contract_on_longer_chain(
```
cducrest: The fact that chain is longer is not relevant any more.
Maybe call it `withdrawable_deposit_contract` or `deposit_contract_on_older_chain` or else ...

schmir-at-bb: right, I actually wanted to fix that. Will open another PR for it later (since I already based some other work on this one).

Issue #63, opened 2019-05-07
Adapt the auction to have min and max participants


cducrest:
related: https://github.com/trustlines-network/project/issues/443

PR #72, opened 2019-05-08
Feature/end auction 50 bidders

cducrest:
Changelog of `ValidatorAuction.sol`:

- change `numberOfParticipants` to `maximalNumberOfParticipants`
- add `minimalNumberOfParticipants`
- change `closingPrice` to `lastBidPrice`
- change auction deployed event to `AuctionDeployed(uint startPrice, uint auctionDurationInDays, uint minimalNumberOfParticipants, uint maximalNumberOfParticipants)`
- change auction ended event to `AuctionEnded(uint closeTime, uint lastBidPrice, uint totalParticipants)`

closes https://github.com/trustlines-protocol/blockchain/issues/63

### DIFF
--- a/contracts/contracts/ValidatorAuction.sol
+++ b/contracts/contracts/ValidatorAuction.sol
@@ -142,18 +142,22 @@ contract ValidatorAuction is Ownable {
 
     function withdraw() public {
         require(auctionState == AuctionState.Ended || auctionState == AuctionState.Failed, "You cannot withdraw before the auction is ended or it failed.");
-        require(bids[msg.sender] > lastBidPrice, "The sender has nothing to withdraw.");
+        require(bids[msg.sender] >= lastBidPrice, "The sender has nothing to withdraw.");
 
-        uint closingPrice;
+        uint valueToWithdraw;
 
         if (auctionState == AuctionState.Ended) {
-            closingPrice = lastBidPrice;
+            valueToWithdraw = bids[msg.sender] - lastBidPrice;
+            assert(valueToWithdraw <= bids[msg.sender]);
+
+            bids[msg.sender] = lastBidPrice;
+        } else {
+            valueToWithdraw = bids[msg.sender];
+            assert(valueToWithdraw <= bids[msg.sender]);
+
+            bids[msg.sender] = 0;
         }
 
-        uint valueToWithdraw = bids[msg.sender] - closingPrice;
-        assert(valueToWithdraw <= bids[msg.sender]);
-
-        bids[msg.sender] = closingPrice;
         msg.sender.transfer(valueToWithdraw);
     }
 
--- a/contracts/contracts/ValidatorAuction.sol
+++ b/contracts/contracts/ValidatorAuction.sol
@@ -157,7 +157,7 @@ contract ValidatorAuction is Ownable {
         msg.sender.transfer(valueToWithdraw);
     }
 
-    function endAuction() internal {
+    function endAuction() internal stateIs(AuctionState.Started) {
         auctionState = AuctionState.DepositPending;
         closeTime = now;
         emit AuctionEnded(closeTime, lastBidPrice, bidders.length);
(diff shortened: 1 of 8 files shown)


contracts/contracts/ValidatorAuction.sol
```diff
@@ -135,7 +142,13 @@ contract ValidatorAuction is Ownable {
 
     function withdraw() public {
         require(auctionState == AuctionState.Ended || auctionState == AuctionState.Failed, "You cannot withdraw before the auction is ended or it failed.");
-        require(bids[msg.sender] > closingPrice, "The sender has nothing to withdraw.");
+        require(bids[msg.sender] > lastBidPrice, "The sender has nothing to withdraw.");
```
berndbohmeier: This contradicts the later code There you compare it to closingPrice, which can be different. 
Now the last bidder can not get his money back if the auction fails. Please add a test for this case. 

To fix this I think we should rewrite the withdraw method to make it more expicit and therefore easier to understand.

what about:
```
assert(bid>= lastBidPrice);
if (auctionstate == Ended) {
    valueToWithdraw = bid - lastBidPrice;
} else if (auctionState == Failed) {
    valueToWithdraw = bid;
}
```

contracts/contracts/ValidatorAuction.sol
```diff
@@ -148,19 +148,32 @@ contract ValidatorAuction is Ownable {
...
+        uint valueToWithdraw = bids[msg.sender] - lowestBidPrice;
+        assert(valueToWithdraw <= bids[msg.sender]);
+
+        bids[msg.sender] = lowestBidPrice;
+
+        msg.sender.transfer(valueToWithdraw);
+    }
+
+    function withdrawAfterAuctionFailed() internal stateIs(AuctionState.Failed) {
+        require(bids[msg.sender] >= 0, "The sender has nothing to withdraw.");
```
berndbohmeier: Here you could no withdraw twice. I would make it symmetric  to the auctionEnded case and use > 0.

contracts/contracts/ValidatorAuction.sol
```diff
@@ -148,19 +148,32 @@ contract ValidatorAuction is Ownable {
...
+        assert(bids[msg.sender] >= lowestBidPrice);
```
berndbohmeier: What does this check?

contracts/contracts/ValidatorAuction.sol
```diff
@@ -148,19 +148,32 @@ contract ValidatorAuction is Ownable {
...
+
+        uint valueToWithdraw = bids[msg.sender];
+        assert(valueToWithdraw <= bids[msg.sender]);
```
berndbohmeier: Do you not trust the compiler? :D In the line above you set it to bid

contracts/contracts/ValidatorAuction.sol
```diff
@@ -44,19 +45,25 @@ contract ValidatorAuction is Ownable {
...
+        uint _minimalNumberOfParticipants,
+        uint _maximalNumberOfParticipants,
         DepositLocker _depositLocker
     ) public
     {
         require(_auctionDurationInDays > 0, "Duration of auction must be greater than 0");
-        require(_numberOfParticipants > 0, "Number of participants must be greater 0");
+        require(_minimalNumberOfParticipants > 0, "Minimal number of participants must be greater than 0");
+        require(_maximalNumberOfParticipants > 0, "Number of participants must be greater than 0");
+        require(_minimalNumberOfParticipants < _maximalNumberOfParticipants, "The minimal number of participants must be smaller than the maximal number of participants.");
```
berndbohmeier: I think we should allow it do be set to the same value, to basically have the old functionality. Does that create problems?

contracts/contracts/ValidatorAuction.sol
```diff
@@ -135,15 +146,44 @@ contract ValidatorAuction is Ownable {
 
     function withdraw() public {
         require(auctionState == AuctionState.Ended || auctionState == AuctionState.Failed, "You cannot withdraw before the auction is ended or it failed.");
-        require(bids[msg.sender] > closingPrice, "The sender has nothing to withdraw.");
+        require(bids[msg.sender] >= lowestBidPrice, "The sender has nothing to withdraw.");
```
berndbohmeier: This is our invariant, so if this fails, we screwed up, so I would say that if it is there this should not be a require but an assert, and the text is wrong. To check if someone can withdraw, we have later the specific checks. 
I would remove this all together.

Issue #69, opened 2019-05-08
auction-deploy status does not work if DepositLocker is not initialized


schmir-at-bb:
auction-deploy status will not work if the DepositLocker is not initialized. It will try to call functions on the slasher contract, which address is zero in that case.

berndbohmeier:
how does the error look like? Is it specific to the 0x address or that the function does not exist? For the later we can not really ensure that it works, as we dont check that the right contract was deployed.

PR #73, opened 2019-05-08
Fix/auction status deposit not init

cducrest:
closes https://github.com/trustlines-protocol/blockchain/issues/69

### DIFF
--- a/auction-deploy/auction_deploy/cli.py
+++ b/auction-deploy/auction_deploy/cli.py
@@ -347,9 +347,11 @@ def status(auction_address, jsonrpc):
         contracts.auction.functions.maximalNumberOfParticipants().call()
     )
     locker_address = contracts.locker.address
-    slasher_address = contracts.slasher.address
+    if contracts.slasher is not None:
+        slasher_address = contracts.slasher.address
     locker_initialized = contracts.locker.functions.initialized().call()
-    slasher_initialized = contracts.slasher.functions.initialized().call()
+    if contracts.slasher is not None:
+        slasher_initialized = contracts.slasher.functions.initialized().call()
 
     # variables
     auction_state_value = contracts.auction.functions.auctionState().call()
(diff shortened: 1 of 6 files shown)


auction-deploy/auction_deploy/core.py
```diff
@@ -59,7 +59,7 @@ def deploy_auction_contracts(
     web3,
     transaction_options: Dict = None,
     private_key=None,
-    auction_options: AuctionOptions,
+    auction_options: AuctionOptions
```
berndbohmeier: something is wrong with your black.

auction-deploy/auction_deploy/cli.py
```diff
@@ -330,7 +330,8 @@ def status(auction_address, jsonrpc):
     locker_address = contracts.locker.address
     slasher_address = contracts.slasher.address
     locker_initialized = contracts.locker.functions.initialized().call()
-    slasher_initialized = contracts.slasher.functions.initialized().call()
+    if locker_initialized:
```
berndbohmeier: This is not the right way to fix this. Why do you check for locker stuff to call then something on slasher. This is you using to much knowledge of how it is implemented. Please fix it by checking stuff on the slasher contract.

auction-deploy/auction_deploy/core.py
cducrest: yes.

auction-deploy/auction_deploy/cli.py
cducrest: I initially had a check that the slasher address is `0`, but this relies on the fact that the address of a contract we did not find is `0`. Is that better?

berndbohmeier: what do you mean we did not find? I though it is 0x if we did not initialized it?

berndbohmeier: I think `get_deployed_auction_contracts` should do those check already, and only set a contract if it has a proper address, otherwise `None` that you can check here if contracts.slasher is None.

PR #74, opened 2019-05-08
Extend validator set contract with historical data

### LINKED ISSUE #62: Store history of finalized changes into the validator set contract
System calls by Parity do not emit events. There we have to store the history of validator set changes to the contracts state.

**Related**
- https://github.com/trustlines-network/project/issues/429

weilbith:
Closes #62.

Please feel free to suggest better names for the variables. Even if they should be informative, they are quite cumbersome...
Also I'm not sure about the functions to retrieve the array lengths. We could make the assumption that non of these arrays will ever contain a zero value and let the user handle the iteration. But I guess that way it is more convenience. I haven't tested them separately. That's because I recognized it does not make much sense, since they need adjustments to the history data, which is already tested. It was just duplicated code. :man_shrugging: 
But I'm open for everything. Just a first proposal.

### DIFF
--- a/contracts/contracts/ValidatorSet.sol
+++ b/contracts/contracts/ValidatorSet.sol
@@ -29,8 +29,8 @@ contract ValidatorSet {
     mapping(address => AddressStatus) status;
     bool public finalized;  // Was the last validator change finalized. Implies currentValidators == pendingValidators
     address public systemAddress = 0xffffFFFfFFffffffffffffffFfFFFfffFFFfFFfE;
-    uint[] public epochStartHeights;
-    mapping(uint => address[]) public epochStartToValidatorsMapping;
+    uint[] epochStartHeights;
+    mapping(uint => address[]) epochValidators;
 
     modifier onlySystem() {
         require(
@@ -69,12 +69,12 @@ contract ValidatorSet {
         return true;
     }
 
-    function getLengthOfEpochStartHeights() external view returns(uint) {
-        return epochStartHeights.length;
+    function getEpochStartHeights() external view returns(uint[]) {
+        return epochStartHeights;
     }
 
-    function getLengthOfEpochStartToValidatorsMapping(uint _blockNumber) external view returns(uint) {
-        return epochStartToValidatorsMapping[_blockNumber].length;
+    function getValidators(uint _epochStart) external view returns(address[]) {
+        return epochValidators[_epochStart];
     }
 
     /**
@@ -138,7 +138,7 @@ contract ValidatorSet {
         currentValidators = pendingValidators;
         finalized = true;
         epochStartHeights.push(block.number);
-        epochStartToValidatorsMapping[block.number] = currentValidators;
+        epochValidators[block.number] = currentValidators;
     }
 
     function removeValidator(address _validator) internal isFinalized returns (bool _success) {
--- a/contracts/contracts/ValidatorSet.sol
+++ b/contracts/contracts/ValidatorSet.sol
@@ -29,6 +29,8 @@ contract ValidatorSet {
     mapping(address => AddressStatus) status;
     bool public finalized;  // Was the last validator change finalized. Implies currentValidators == pendingValidators
     address public systemAddress = 0xffffFFFfFFffffffffffffffFfFFFfffFFFfFFfE;
+    uint[] public blockNumbersOfFinalizedChanges;
+    mapping(uint => address[]) public finalizedValidatorSets;
 
     modifier onlySystem() {
         require(
@@ -65,6 +67,14 @@ contract ValidatorSet {
         finalized = true;
         initiated = true;
         return true;
+    }
+
+    function getLengthOfBlockNumbersOfFinalizedChanges() external view returns(uint) {
+        return blockNumbersOfFinalizedChanges.length;
+    }
+
+    function getLengthOfFinalizedValidatorSet(uint _blockNumber) external view returns(uint) {
+        return finalizedValidatorSets[_blockNumber].length;
     }
 
     /**
@@ -127,6 +137,8 @@ contract ValidatorSet {
     function finalizeChange() public onlySystem {
         currentValidators = pendingValidators;
         finalized = true;
+        blockNumbersOfFinalizedChanges.push(block.number);
+        finalizedValidatorSets[block.number] = currentValidators;
     }
 
     function removeValidator(address _validator) internal isFinalized returns (bool _success) {
--- a/contracts/contracts/ValidatorSet.sol
+++ b/contracts/contracts/ValidatorSet.sol
@@ -29,8 +29,8 @@ contract ValidatorSet {
     mapping(address => AddressStatus) status;
     bool public finalized;  // Was the last validator change finalized. Implies currentValidators == pendingValidators
     address public systemAddress = 0xffffFFFfFFffffffffffffffFfFFFfffFFFfFFfE;
-    uint[] public blockNumbersOfFinalizedChanges;
-    mapping(uint => address[]) public finalizedValidatorSets;
+    uint[] public epochStartHeights;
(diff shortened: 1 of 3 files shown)


contracts/tests/test_validator_set.py
```diff
@@ -142,6 +142,49 @@ def test_cannot_call_finalize_change(validator_set_contract_session, accounts):
         contract.functions.finalizeChange().transact({"from": accounts[0]})
 
 
+def test_finalize_change_update_history_data_correctly(
```
jannikluhn: Probably better to split this up into multiple tests, each testing one state transition.

contracts/tests/test_validator_set.py
```diff
@@ -142,6 +142,49 @@ def test_cannot_call_finalize_change(validator_set_contract_session, accounts):
...
+):
+    assert (
+        validator_set_contract_session.functions.getLengthOfBlockNumbersOfFinalizedChanges().call()
+        == 0
+    )
+
+    new_validator_set = accounts[:2]
+    validator_set_contract_session.functions.testChangeValiatorSet(
+        new_validator_set
+    ).transact()
```
jannikluhn: We should probably test that nothing has changed at this point.

contracts/tests/test_validator_set.py
```diff
@@ -142,6 +142,49 @@ def test_cannot_call_finalize_change(validator_set_contract_session, accounts):
         contract.functions.finalizeChange().transact({"from": accounts[0]})
 
 
+def test_finalize_change_update_history_data_correctly(
+    validator_set_contract_session, accounts, web3
```
jannikluhn: Do I understand correctly that `validator_set_contract_session` carries state between tests? If so I don't think this is a good idea. Why not just make this a single-test scoped fixture?

berndbohmeier:
Did you check that everything works fine if you used the changed contract to handle validators on a parity?

weilbith:
@jannikluhn I like the name suggestion! :blush: 

@berndbohmeier yes, I have and it works. It gets applied and I can query the history data. After the first time applied, I see the first entry with the finalized block number and the address list. Then reporting one of two validators which gets applied extends this history.

contracts/tests/test_validator_set.py
weilbith: Oh, I wasn't aware about it. To be honest I haven't expected it for such a contract. I removed it at the fixture and all tests are still running.

contracts/tests/test_validator_set.py
weilbith: My proposal is the following split:
- check that finalization is necessary (noted by you [here](https://github.com/trustlines-protocol/blockchain/pull/74#discussion_r282150530))
- check if new epoch start height has been added to history
- check if new validator set has been added to history

Enough or even more cases?

contracts/tests/test_validator_set.py
weilbith: I have added a test case for that. :+1:

contracts/contracts/ValidatorSet.sol
```diff
@@ -67,6 +69,14 @@ contract ValidatorSet {
         return true;
     }
 
+    function getLengthOfEpochStartHeights() external view returns(uint) {
```
cducrest: Is this the number of epochs? If so, call it `numberOfEpcohs` or `getNumberOfEpochs`.

weilbith: Yeah, this ugly name was for sticking to the name of the variable. But related to trustlines-protocol/blockchain#74 we probably provide much more comfortable functions and make the variables not public.

contracts/contracts/ValidatorSet.sol
```diff
@@ -29,6 +29,8 @@ contract ValidatorSet {
     mapping(address => AddressStatus) status;
     bool public finalized;  // Was the last validator change finalized. Implies currentValidators == pendingValidators
     address public systemAddress = 0xffffFFFfFFffffffffffffffFfFFFfffFFFfFFfE;
+    uint[] public epochStartHeights;
+    mapping(uint => address[]) public epochStartToValidatorsMapping;
```
cducrest: you could name this `epochValidators` or `validatorsOfEpoch`

contracts/contracts/ValidatorSet.sol
```diff
@@ -67,6 +69,14 @@ contract ValidatorSet {
         return true;
     }
 
+    function getLengthOfEpochStartHeights() external view returns(uint) {
+        return epochStartHeights.length;
+    }
+
+    function getLengthOfEpochStartToValidatorsMapping(uint _blockNumber) external view returns(uint) {
```
cducrest: Is this the number of validators of a given epoch?
If so, shall it be called `numberOfValidatorsInEpoch`

contracts/contracts/ValidatorSet.sol
weilbith: So I go for `getEpochStartHeights()` and `getValidators(uint _epochStart)`. Your thoughts?

contracts/contracts/ValidatorSet.sol
weilbith: Okay, if you already suggest I take the first one. I like it when variables which are somehow related have the same prefix. :grin:

contracts/contracts/ValidatorSet.sol
weilbith: See the [conversation earlier](https://github.com/trustlines-protocol/blockchain/pull/74#discussion_r282804941).

contracts/contracts/ValidatorSet.sol
cducrest: `getEpochStartHeights()` is incorrect, you don't get the height of an epoch start, you get the number of epochs. So `getNumberOfEpochs` seems to be the optimal name for me.

weilbith: No, I changes it to 
```solidity
function getEpochStartHeights() external view returns(uint[]) {
    return epochStartHeights;
}
```
 Still think it is incorrect?

weilbith: Checkout the last commit. Else I overwrite it.

contracts/tests/test_validator_set.py
weilbith: Test code has been simplified much by the new interface.

contracts/contracts/ValidatorSet.sol
cducrest: Sure, if you change what the function does, now it is fine :).

contracts/tests/conftest.py
```diff
@@ -72,7 +72,7 @@ def validators(accounts, malicious_validator_address):
     return accounts[:HONEST_VALIDATOR_COUNT] + [malicious_validator_address]
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture()
 def validator_set_contract_session(deploy_contract, validators, web3):
```
jannikluhn:
```suggestion
def validator_set_contract(deploy_contract, validators, web3):
```

as I assume `session` refers to the scope

cducrest: This fixture is used in between every tests: https://github.com/trustlines-protocol/contract-deploy-tools/blob/master/deploy_tools/plugin.py#L107

So please use the scope `session` when deploying contracts so that the tests do not redeploy the contract each time. This will speed up testing.

weilbith: True. Solved. :)

weilbith: Oh. There is a reason because you should touch already approved code with care...

weilbith: @come I rebased the branch on itself and removed the change. Would love to see an approval, before I merge. Thanks for all the input. :v:

Issue #65, opened 2019-05-07
Change cli to parse date for locker


cducrest:
to deploy auction via cli, have possibility to use a date instead of timestamp.
related: https://github.com/trustlines-network/project/issues/445

PR #76, opened 2019-05-09
Feature/issue 65 change cli to parse date for locker

schmir-at-bb:
This allows the deposit lockers release date to be set via the --release-date option and implements #65 

It contains some additional changes to show an actual date instead of timestamps when running the status command.

### DIFF
--- a/auction-deploy/auction_deploy/cli.py
+++ b/auction-deploy/auction_deploy/cli.py
@@ -1,6 +1,7 @@
 from enum import Enum
 
 import click
+import pendulum
 from web3 import Web3, EthereumTesterProvider, Account
 from deploy_tools.deploy import send_function_call_transaction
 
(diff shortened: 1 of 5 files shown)


auction-deploy/auction_deploy/cli.py
```diff
@@ -38,6 +39,17 @@ def validate_address(ctx, param, value):
...
 
 
+def validate_date(ctx, param, value):
+    if value is None:
+        return None
+    try:
+        return pendulum.parse(value)
+    except pendulum.parsing.exceptions.ParserError as e:
+        raise click.BadParameter(
+            f'The parameter "{value}" cannot be parsed as a date'
```
berndbohmeier: Can we maybe add some information on the expected format?

auction-deploy/auction_deploy/cli.py
```diff
@@ -147,7 +159,16 @@ def main():
     "release_timestamp",
     help="The release timestamp of the deposit locker",
     type=int,
-    required=True,
+    required=False,
+)
+@click.option(
+    "--release-date",
+    "release_date",
+    help="The release date of the deposit locker",
```
berndbohmeier: Can we maybe add some info about the expected format?

PR #79, opened 2019-05-10
Remove the DepositPending state

berndbohmeier:
There is the weirdness that endAuction will fire the event AuctionEnded
but change to the state DepositPending. One solution to fix this is to
remove the intermediate state DepositPending. This has the advantage
that people can start withdrawing right after the auction has ended.
One might however argue that this is less safe. Let me know what you think.


berndbohmeier:
Rejected during discussion, as it makes the logic more complex.

PR #80, opened 2019-05-13
Add AuctionDepositPending event

berndbohmeier:
Add AuctionDepositPending event which is emitted instead of AuctionEnded
to avoid the confusion that the state(DepositPending) is different from
the event (AuctionEnded)
The AuctionEnded event is emitted later when transitioning to the
AuctionEnded state with the same parameters.


berndbohmeier:
Woops there a fix to the Ownable contract in there, this was not on purpose. I can remove it if you want, but I think we need it anyways. I will fix this upstream.

PR #82, opened 2019-05-13
Add printing of the addresses within the contracts to check them.

cducrest:


auction-deploy/auction_deploy/cli.py
```diff
@@ -464,6 +475,47 @@ def status(auction_address, jsonrpc):
...
+        click.echo(
+            "The locker address in the auction contract corresponds to the locker address"
+        )
+    else:
+        click.secho(
+            "The locker address in the auction contract does not match to the locker address",
+            fg="red",
+        )
+    if auction_address_in_locker_contract == auction_address:
+        click.echo(
```
berndbohmeier: Why do we need to write that? This should be the case right? I would maybe only show the warning if something is wrong, otherwise the output is to much and someone will start ignoring parts of it anyways.

cducrest: It should be the case, but I thought it is one of the things we should check before starting the auction.
I removed the happy case printing.

Issue #83, opened 2019-05-17
Try upgrading solidity to 0.5.7 and collect what needs to be changed. 


berndbohmeier:

schmir-at-bb:
The referenced PRs above should be  nearly everything that needs to be changed.

schmir-at-bb:
The changelog makes me feel like we should stay on 0.4.x: https://github.com/ethereum/solidity/blob/develop/Changelog.md

berndbohmeier:
I fail to see what you changed, you only where forced to make stuff explicit, no?
So I would think we should update.

schmir-at-bb:
> I fail to see what you changed, you only where forced to make stuff explicit, no?

Yes. The commit messages are fairly detailed.You should probably read them. 

> So I would think we should update.

Well, the changes being small was a requirement for us to do it. That should not mean that we have to do it. After reading the solidity changelog, I also feel like we should use the latest released version if we switch to 0.5.x. (0.5.8 does contain some important bugfixes).

berndbohmeier:
I read them. Making the data location explicit is not a change for me. The data location was the same, only you now have to declare it.

schmir-at-bb:
> I read them. 

sorry, for implying the opposite.

> Making the data location explicit is not a change for me. The data location was the same, only you now have to declare it

yes, I agree. The real change is the different compiler version being used.

(as an aside I remember spending at least a day on a bug in gcc I hit, when I changed the type of a parameter from `double` to `const double`. That was 18 years ago. Not sure what that tells us.Probably that I'm old)

schmir-at-bb:
we changed the builder image and contract-deploy-tools to support solidity 0.5.8. A PR is in review that changes the solidity version to 0.5.8 in the blockchain repo.

PR #85, opened 2019-05-17
adapt code to compile with solidity 0.5.8

schmir-at-bb:
The code changes needed are mostly explicitly setting the data location. I've
followed what the solidity compiler was already suggesting. Breaking changes in
to 0.5.x series are documented in [1].

The problem is, that a lot of tests are failing now, some of them with something
like 'eth.exceptions.InvalidInstruction: Invalid opcode 0x1c @ 18'. My guess
here is that the new default target option is to blame. It changed from
byzantium in 0.5.4 [2] to petersberg in 0.5.7 [3]. My guess is that the default
in 0.4.25 is byzantium.

See https://github.com/trustlines-protocol/blockchain/issues/83

[1] https://solidity.readthedocs.io/en/latest/050-breaking-changes.html
[2] https://solidity.readthedocs.io/en/v0.5.4/using-the-compiler.html#target-options
[3] https://solidity.readthedocs.io/en/v0.5.7/using-the-compiler.html#target-options

### DIFF
--- a/auction-deploy/requirements.txt
+++ b/auction-deploy/requirements.txt
@@ -1,8 +1,7 @@
 click
 web3
 pendulum
-git+git://github.com/trustlines-protocol/contract-deploy-tools.git
-
+contract-deploy-tools
 setuptools_scm
 pytest
 flake8
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,9 +3,9 @@ version: 2.1
 executors:
   ubuntu-builder:
     docker:
-      - image: trustlines/builder:master8
+      - image: trustlines/builder:master20
         environment:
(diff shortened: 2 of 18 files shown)


schmir-at-bb:
I think this is now ready to be merged.

auction-deploy/requirements.txt
```diff
@@ -1,8 +1,7 @@
 click
 web3
 pendulum
-git+git://github.com/trustlines-protocol/contract-deploy-tools.git
-
```
berndbohmeier: I think they were separated on purpose based on runtime dependencies and develop dependencies, so please give us back the blank line.

berndbohmeier:
reopend in #118 because I can not change things here

Issue #88, opened 2019-05-20
Create list of bootstrap nodes


jannikluhn:

schmir-at-bb:
merged as of 122c3c1238b33eff92ad0e094b25d1d4

PR #98, opened 2019-05-21
Feature/extract bootnodes

schmir-at-bb:
Implements part of #88, also see  https://github.com/trustlines-network/project/issues/479


config/extract-bootnodes
```diff
@@ -0,0 +1,71 @@
...
+
+
+def main():
+    if len(sys.argv) != 2:
+        print("Usage: extract-bootnodes /path/to/nodes.json")
+        sys.exit(1)
+
+    nodes_json_path = sys.argv[1]
+    nodes = read_nodes_json(nodes_json_path)
+    nodes = [node for node in nodes if node["last_contact"].get("success", 0)]
```
weilbith: I tested it with all chains I have on my machine ever synced. Some map `"last_contact": null`. Therefore I get a `AttributeError` for the `NoneType`. In the end I don't get any nodes. I would say we should check this and ignore these nodes.

Issue #107, opened 2019-05-23
Deploy script for reward contract


berndbohmeier:
This needs to be within the same command as the deploy bridge contract on tlbc because they need to know each others address.

saschagoebel:
Already done as part of #104

PR #119, opened 2019-05-24
Add reward contract

### LINKED ISSUE #106: Create reward contract
For necessary changes, have a look at https://github.com/trustlines-network/project/wiki/PoA-Network-Bridge

jannikluhn:
Closes #106 

- The contract has been copied from https://github.com/poanetwork/poa-network-consensus-contracts@master.
- Emission funds are set to zero as we don't use it.
- The [_isMiningActive check](https://github.com/poanetwork/poa-network-consensus-contracts/blob/master/contracts/RewardByBlock.sol#L76) was removed as we did in earlier tests (I'm quite sure it's unnecessary)
- The contract has been simplified by removing the dependency on an external keys manager contract. This is new, but justified as keeping it in there without actually deploying the contract is a hack that just works by accident.
- I added a GPL license file in the subdirectory and wrote a short readme with notable changes which I think is required by GPL.

I haven't tried to compile the contracts. This is left for #107


andfletcher:
We need to put the bridge into a different repo since the repo we have is MIT and the poa bridge is GPL 3. Basically, if the sub-repos have a different license then the main repo it is very confusing since normally the license of the repo dominates over the sub repos.

schmir-at-bb:
> We need to put the bridge into a different repo since the repo we have is MIT and the poa bridge is GPL 3. 

This could easily be solved by documenting it instead. We've wasted lots of time with stuff like that already and we should at some point stop doing it. This is annoying.

jannikluhn:
I'd also rather keep them in the same repo, to make sharing of contract addresses easier. I can see though why it could be confusing right now.

Should we maybe have a single license file in the root and put something like "everything in `bridge/` is GPL lincensed, everything else is MIT in there?

andfletcher:
> I'd also rather keep them in the same repo, to make sharing of contract addresses easier. I can see though why it could be confusing right now.
> 
> Should we maybe have a single license file in the root and put something like "everything in `bridge/` is GPL lincensed, everything else is MIT in there?

I think that would work but we need to remove the license of the repo then (since it says that it is MIT in the top right corner)

jannikluhn:
I've removed the license file in the `bridge/reward-contract` subdirectory, put both MIT and GPL into the main license file, and wrote a short sentence about which license is supposed to apply to which part of the repository. I assume that Github will detect that the license file has changed and update the indicator accordingly once we merge this PR. That should suffice, no?

andfletcher:
yes but i think you should double check if the license has changed after merge

jannikluhn:
> yes but i think you should double check if the license has changed after merge

Confirmed that it has changed: Now it's just a button "show license" which leads to our custom license file.

PR #120, opened 2019-05-24
Feature/bridge docker

### LINKED ISSUE #108: Create docker compose file to start bridge workers
This does not have to include the mainchain and tlbc nodes. 
Config files for the bridge need to be adjusted to our needs.

jannikluhn:
Closes #108 

This adds a docker compose file and a `.env` file based on the examples from the POA token bridge repo. The repo itself is added as a submodule because it is used by the docker compose file to build the containers.

The `.env` file is based on `.env.example` with some unnecessary lines removed. Details that still have to be filled in are marked with `TODO` (e.g. contract addresses).

The URLs of the (centralized) gas price oracle have been removed based on my understanding that in this case the gas price is queried from the contract (where the validators can vote on it).

In the docker-compose file, only some paths are changed. Notably, we still start some containers such as `foreign_sender` which are likely not necessary, but I have no way to confirm this right now and they don't hurt anyone.

It has been confirmed that it's possible to start up the docker containers, but more than this it not really possible at this stage (no nodes, no deployed contracts).


jannikluhn:
We have to include the repo somehow as the docker compose file needs it to build the image. I guess we could also just copy it, but I guess we'll want to update it from time to time, so that's why I thought a submodule makes sense. What's the downside?

schmir-at-bb:
from the top of my head:
- everyone needs to learn how to work with submodules
- you need to remember to run a new command after checking out the git repo
- the repo itself is not self-contained anymore
- changes are hard to see when updating the submodule

git subtree would be an alternative, that doesn't require everyone to learn a new command.
see https://www.atlassian.com/blog/git/alternatives-to-git-submodule-git-subtree

How many files are we talking about?

schmir-at-bb:
> We have to include the repo somehow as the docker compose file needs it to build the image

Another alternative would be to build the image ourselves and upload it to docker.

jannikluhn:
I realized docker compose can also build directly from Github. So I removed the submodule and updated the docker compose file accordingly. If we need to change something in the future, we can just fork the repo and update the URL.

Issue #94, opened 2019-05-20
Update chain spec file to use validator set contract at communicated block height


jannikluhn:

cducrest:
Desired block number of fork: 2717171
Timestamp of block 2565494: 1558713735
average block time: 5.59s
`(2717171 - 2565494) * 5.59 + 1558713735 = 1559561609`

This timestamp corresponds to Monday the third of June at approx 13:30 with current minting speed

PR #122, opened 2019-05-24
Add fork at block height 2717171 for the validator set contract

cducrest:
Closes https://github.com/trustlines-protocol/blockchain/issues/94
Please check the validator contract before accepting / merging, potentially using the tool brought in: https://github.com/trustlines-protocol/blockchain/pull/121


berndbohmeier:
I checked the validator addresses and the contract address.

PR #124, opened 2019-05-27
Feature/erc20 token

cducrest:
close https://github.com/trustlines-network/project/issues/354
Functions rely on checks by safeMath.lib, we could make requires more obvious and remove safeMath.lib
Otherwise I'd wish to refactor safeMath.lib in contract-library to use a more recent version with a tracked source.

### DIFF
--- a/contracts/contracts/TrustlinesNetworkToken.sol
+++ b/contracts/contracts/TrustlinesNetworkToken.sol
@@ -4,8 +4,10 @@ import "./SafeMath.sol";
 
 
 contract TrustlinesNetworkToken {
+
     using SafeMath for uint256;
 
+    uint constant MAX_UINT = 2**256 - 1;
     string private _name;
     string private _symbol;
     uint8 private _decimals;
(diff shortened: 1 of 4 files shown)


contracts/contracts/TrustlinesNetworkToken.sol
```diff
@@ -0,0 +1,126 @@
...
+        _approve(msg.sender, spender, value);
+    }
+
+    function transferFrom(address sender, address recipient, uint256 amount) public returns (bool) {
+        _transfer(sender, recipient, amount);
+
+        uint allowance = _allowances[sender][msg.sender];
+        uint updatedAllowance = allowance.sub(amount);
+        if (allowance < MAX_UINT) {
+            _approve(sender, msg.sender, updatedAllowance);
```
berndbohmeier: why do you reuse `_approve` here. This seems wrong, as you dont approve again here, so just substract from the allowance.

cducrest: That's just how they do it in the original: https://github.com/OpenZeppelin/openzeppelin-solidity/blob/master/contracts/token/ERC20/ERC20.sol#L96
I'm fine with changing it, I just changed the minimum I could not to introduce complexity / bugs.

cducrest: But I also want to throw an approve event, otherwise we cannot track the change of allowance through events.

berndbohmeier: I am fine with it

PR #143, opened 2019-06-04
Upgrade rlp reader to newer version

### LINKED ISSUE #142: Fix RLPReader.sol / Equivocation inspector
for fixes, see https://github.com/trustlines-network/project/issues/492
Also, see adjustments made to the source contract: https://github.com/hamdiallam/Solidity-RLP/blob/master/contracts/RLPReader.sol#L66

berndbohmeier:
Upgrade the rlp reader lib to a newer version to fix a potentially
security error in the lib when reading from an empty byte string.
To be more save, we also revert on empty bytes.

Closes #142

### DIFF
--- a/contracts/contracts/lib/RLPReader.sol
+++ b/contracts/contracts/lib/RLPReader.sol
@@ -11,9 +11,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * 
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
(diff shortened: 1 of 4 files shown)


contracts/contracts/lib/RLPReader.sol
```diff
@@ -188,26 +209,49 @@ library RLPReader {
...
+
         uint offset = _payloadOffset(item.memPtr);
         uint len = item.len - offset;
+
+        uint result;
         uint memPtr = item.memPtr + offset;
+        assembly {
+            result := mload(memPtr)
+
+        // shfit to the correct location if neccesary
```
schmir-at-bb: s/shfit/shift/

PR #144, opened 2019-06-04
Add ProxyValidatorContract

cducrest:
it is currently not used by the ValidatorSet contract but just serves as first step towards getting validators to the bridge

### DIFF
--- a/contracts/contracts/ValidatorProxy.sol
+++ b/contracts/contracts/ValidatorProxy.sol
@@ -1,0 +1,31 @@
+pragma solidity ^0.5.8;
+
+
+contract ValidatorProxy {
+
+    mapping (address => bool) public isValidator;
+    address public systemAddress = 0xffffFFFfFFffffffffffffffFfFFFfffFFFfFFfE;
+    address[] public validators;
+
+    function updateValidators(address[] memory newValidators) public {
+        require(tx.origin == systemAddress, "Only the system address can be responsible for the call of this function.");  // solium-disable-line security/no-tx-origin
+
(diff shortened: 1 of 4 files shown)


contracts/contracts/ValidatorProxy.sol
```diff
@@ -0,0 +1,31 @@
+pragma solidity ^0.5.8;
+
+
+contract ValidatorProxy {
+
+    mapping (address => bool) public isValidator;
+    address public systemAddress = 0xffffFFFfFFffffffffffffffFfFFFfffFFFfFFfE;
+    address[] public validators;
+
+    function updateValidators(address[] memory newValidators) public {
```
saschagoebel: Do we have to support this function in the proxy? Or is this just for testing purposes?

contracts/contracts/ValidatorProxy.sol
```diff
@@ -0,0 +1,31 @@
+pragma solidity ^0.5.8;
+
+
+contract ValidatorProxy {
```
saschagoebel: I think the proxy should follow this interface, right? https://github.com/poanetwork/poa-bridge-contracts/blob/master/contracts/IBridgeValidators.sol

contracts/contracts/ValidatorProxy.sol
cducrest: This is the proxy to the ValidatorSet that allows to have access to the set of validators on the chain even though the active ValidatorSet contract might change. This is not related to the bridge.

saschagoebel:
Already merged im #150

PR #149, opened 2019-06-05
Bridge E2E test

jannikluhn:
It starts the containers, two nodes, checks that the containers are running, and stops. That's all we can do for now until we can deploy the contracts.

The containers running test isn't working because I didn't want to waste too much time writing cryptic bash code I don't understand and it's just a placeholder for now anyways.


weilbith:
I rebased this onto master, because we need the recent updates of the deploy tools.
Furthermore I have extended the configurations in preparation for contracts deployment.
The current script which is work in progress, can be checked out [here](https://github.com/weilbith/blockchain/blob/feature/bridge-e2e/bridge/e2e_tests/run_e2e_test.sh). This branch is used to exchange development results which are not ready for the actual PR here. Nevertheless this script has evolved a lot already. It uses the tlbc-monitor e2e tests as example.

PR #151, opened 2019-06-06
Bridge home contract deployment

### LINKED ISSUE #104: Deploy script for bridge contract on tlbc

weilbith:
Closes #104


bridge-deploy/Makefile
```diff
@@ -0,0 +1,41 @@
+VIRTUAL_ENV ?= venv
+
+lint: install-requirements
+	$(VIRTUAL_ENV)/bin/flake8 bridge_deploy tests setup.py
+	$(VIRTUAL_ENV)/bin/black --check bridge_deploy tests setup.py
+	$(VIRTUAL_ENV)/bin/mypy bridge_deploy tests setup.py --ignore-missing-imports
+
+format:
```
cducrest: I guess this should be `.PHONY`

bridge-deploy/Makefile
```diff
@@ -0,0 +1,41 @@
...
+	$(VIRTUAL_ENV)/bin/black bridge_deploy tests setup.py
+
+test: install-requirements compile
+	$(VIRTUAL_ENV)/bin/pytest -s tests
```
cducrest: Should remove the `-s` otherwise pytest will print confusing stuff from non failing tests. (afaik)

bridge-deploy/bridge_deploy/cli.py
```diff
@@ -0,0 +1,276 @@
...
+        return validate_and_format_address(value)
+    except InvalidAddressException as e:
+        raise click.BadParameter(
+            f"The address parameter is not recognized to be an address: {value}"
+        ) from e
+
+
+validator_set_address_option = click.option(
+    "--validator-set-address",
+    help='The address of the validator set contract, "0x" prefixed string',
```
cducrest: Is it? 
I thought the address given here should be either the address of the proxy or of the bridge validators (probably the proxy).

bridge-deploy/bridge_deploy/cli.py
```diff
@@ -0,0 +1,276 @@
...
+    "--block-reward-address",
+    help='The address of the block reward contract, "0x" prefixed string',
+    type=str,
+    required=True,
+    callback=validate_address,
+    metavar="BLOCK_REWARD_ADDRESS",
+    envvar="BLOCK_REWARD_ADDRESS",
+)
+
+owner_address_option = click.option(
```
cducrest: What is the owner's power for the bridge? Shouldn't we use `0x00...01` since we cannot have governance of the bridge (and 0x00...0 is not allowed).

bridge-deploy/bridge_deploy/cli.py
```diff
@@ -0,0 +1,276 @@
...
+    private_key = retrieve_private_key(keystore)
+
+    nonce = get_nonce(
+        web3=web3, nonce=nonce, auto_nonce=auto_nonce, private_key=private_key
+    )
+    transaction_options = build_transaction_options(
+        gas=gas, gas_price=gas_price, nonce=nonce
+    )
+
+    dir(transaction_options)
```
cducrest: What is this for?

bridge-deploy/bridge_deploy/cli.py
```diff
@@ -0,0 +1,276 @@
...
+        validator_contract_address=validator_set_address,
+        home_daily_limit=home_daily_limit,
+        home_max_per_tx=home_max_per_tx,
+        home_min_per_tx=home_min_per_tx,
+        home_gas_price=home_gas_price,
+        required_block_confirmations=required_block_confirmations,
+        block_reward_address=block_reward_address,
+        foreign_daily_limit=foreign_daily_limit,
+        foreign_max_per_tx=foreign_max_per_tx,
+        home_bridge_owner=owner_address,
```
cducrest: You can probably pack all the `init` arguments in one variable if that improves readability.

bridge-deploy/bridge_deploy/home.py
```diff
@@ -0,0 +1,118 @@
...
+    # Use proxy from now on
+    home_bridge_contract = web3.eth.contract(
+        abi=home_bridge_src["abi"], address=home_bridge_storage_contract.address
+    )
+
+    # TODO: Initialize bridge
+
+    # TODO: transfer ownership
+
+    contracts = DeployedHomeBridgeResult(home_bridge_contract, latest_block.number)
```
cducrest: I still believe it is weird to return the block at which the contract was deployed, and do not see where you use it except for printing it.

bridge-deploy/tests/test_cli.py
```diff
@@ -0,0 +1,48 @@
...
+
+    # TODO: Add actual working addresses and contracts
+    result = runner.invoke(
+        main,
+        args=(
+            "deploy-home"
+            " --jsonrpc test "
+            " --validator-set-address 0x0000000000000000000000000000000000000000"
+            " --block-reward-address 0x0000000000000000000000000000000000000000"
+            " --owner-address 0x0000000000000000000000000000000000000000"
```
cducrest: Shouldn't this fail with owner = addres(0) ?
https://github.com/poanetwork/poa-bridge-contracts/blob/master/contracts/upgradeable_contracts/erc20_to_native/HomeBridgeErcToNative.sol#L142

bridge-deploy/bridge_deploy/cli.py
```diff
@@ -0,0 +1,276 @@
...
+)
+
+required_block_confirmations_option = click.option(
+    "--required-block-confirmations",
+    help="The number of blocks to wait for finality",
+    type=int,
+    default=4,
+)
+
+foreign_daily_limit_option = click.option(
```
jannikluhn: I don't think we need this option as the limit should either not exist (if we write our own burning bridge contract) or be set to zero.

bridge-deploy/bridge_deploy/cli.py
```diff
@@ -0,0 +1,276 @@
...
+    "--foreign-daily-limit",
+    help="The daily transfer limit for the foreign bridge in WEI",
+    type=int,
+    default=15_000_000_000_000_000_000_000_000,
+)
+
+foreign_max_per_tx_option = click.option(
```
jannikluhn: ditto

bridge-deploy/bridge_deploy/home.py
```diff
@@ -0,0 +1,118 @@
...
+):
+
+    if transaction_options is None:
+        transaction_options = {}
+
+    latest_block = web3.eth.getBlock("latest")
+
+    # Deploy home bridge proxy
+    eternal_storage_proxy_src = load_contract("EternalStorageProxy")
+    home_bridge_storage_contract: Contract = deploy_compiled_contract(
```
jannikluhn: why do we need the type hint here? Is it missing in the deploy tools?

bridge-deploy/bridge_deploy/home.py
jannikluhn: The bridge workers need it to limit the number of blocks in which they look for events.

bridge-deploy/bridge_deploy/cli.py
jannikluhn: Agree that we don't need the option. I think `0` is only not allowed during initialization, we can still set it afterwards. Would prefer having a temporary owner and setting it to `0` later over setting it to `1` during initialization -- mostly for aesthetic purposes.

bridge-deploy/bridge_deploy/home.py
```diff
@@ -0,0 +1,118 @@
...
+        validator_contract_address,
+        home_daily_limit,  # 30000000000000000000000000,  # HOME_DAILY_LIMIT in WEI,
+        home_max_per_tx,  # 1500000000000000000000000,  # HOME_MAX_AMOUNT_PER_TX in WEI,
+        home_min_per_tx,  # 500000000000000000,  # HOME_MIN_AMOUNT_PER_TX in WEI,
+        home_gas_price,  # 1000000000,  # HOME_GAS_PRICE in WEI,
+        required_block_confirmations,  # 4,  # HOME_REQUIRED_BLOCK_CONFIRMATIONS,
+        block_reward_address,  # "0x0000000000000000000000000000000000000000",  # BLOCK_REWARD_ADDRESS,
+        foreign_daily_limit,  # 15000000000000000000000000,  # FOREIGN_DAILY_LIMIT,
+        foreign_max_per_tx,  # 750000000000000000000000,  # FOREIGN_MAX_AMOUNT_PER_TX,
+        home_bridge_owner,  # "0x0000000000000000000000000000000000000000",  # HOME_BRIDGE_OWNER
```
jannikluhn: see above, I think this should be set to the private key's address and then in another transaction to 0.

bridge-deploy/setup.py
```diff
@@ -0,0 +1,15 @@
+from setuptools import setup, find_packages
+
+setup(
+    name="bridge-deploy",
+    setup_requires=["setuptools_scm"],
+    # use_scm_version=True,
+    version="0.0.1",
+    packages=find_packages(),
+    package_data={"bridge_deploy": ["contracts.json"]},
+    install_requires=["click", "web3", "contract-deploy-tools", "pendulum"],
```
jannikluhn: What do we need pendulum for? Can't see it referenced anywhere

bridge-deploy/Makefile
saschagoebel: True

bridge-deploy/Makefile
saschagoebel: Debugging leftover ;-)

bridge-deploy/bridge_deploy/cli.py
saschagoebel: Could be both. Clarified description.

bridge-deploy/bridge_deploy/cli.py
saschagoebel: That way we could never change anything regarding the bridge later on. This could also be a MultiSigWallet?

bridge-deploy/bridge_deploy/cli.py
saschagoebel: Dev-Leftover. Removed.

bridge-deploy/bridge_deploy/cli.py
saschagoebel: How would that look like? Like a Dict? Are the parameters still required then?

bridge-deploy/bridge_deploy/home.py
saschagoebel: Exactly. We could get it from the bridge contract itself later on, I think, but that would be even more complicated.

bridge-deploy/tests/test_cli.py
saschagoebel: Yes. It does.
We still need to hook this up to the validator set proxy and block reward contract.
These need to be merged / deployed first.
Are we going to deploy the reward contract via the chainspec?

bridge-deploy/bridge_deploy/cli.py
saschagoebel: 0 is not allowed by the contract, but we can set it to 1 WEI.
I wanted to make it configurable, but we can also hard-code the foreign options.

saschagoebel: Changed.

bridge-deploy/bridge_deploy/cli.py
saschagoebel: Changed.

bridge-deploy/bridge_deploy/home.py
saschagoebel: Honestly: Copy-Pasted from Auction-Deploy. Seems not to be required. Removed.

bridge-deploy/bridge_deploy/cli.py
jannikluhn: Only safe option would be to have a multisig consisting of the validators. I guess we could implement this in the validator proxy, but it would take some effort. So I think we should just pick some good defaults, keep them static, and update it by forking if necessary (I guess this requires the update-contract-state-in-fork feature in Parity though).

bridge-deploy/bridge_deploy/cli.py
jannikluhn: It's only restricted in `initialize`, after that the owner can set it to 0.

bridge-deploy/bridge_deploy/home.py
saschagoebel: Initializing with deployer for now.

bridge-deploy/setup.py
saschagoebel: Missed that one when removing it from the dependencies.

bridge-deploy/bridge_deploy/cli.py
saschagoebel: Set the tx limit to 0 and the daily limit to 1 for now. Will update with a separate tx.

bridge-deploy/bridge_deploy/cli.py
saschagoebel: setOwner is internal, but the method we should actually use is transferProxyOwnership anyway, which again checks for address(0). Resorting to setting the owner to address(1) now.

bridge-deploy/tests/test_cli.py
saschagoebel: Changed to address(2) for now, but the test still fails. Maybe a builtin is not considered a valid contract. I'll investigate.

bridge-deploy/bridge_deploy/home.py
saschagoebel: Changed to initializing with address(1), as changing the owner to address(0) later seems to be forbidden as well.

bridge-deploy/bridge_deploy/cli.py
cducrest: I would say like a `NamedeTuple` similarly to `transaction_options`
By init arguments I just mean that `validator_contract_address, home_daily_limit, home_max_per_tx, ..., block_reward_address` are essentially all arguements for initialize function of the contract, so they represent one input argument of the `initialize_home_bridge_contract` python function. Not the `web3`, `transactions_options`, etc ...

bridge-deploy/bridge_deploy/home.py
cducrest: But you can just use `web3.eth.blockNumber` before calling initialize, it doesn't matter to be incorrect at +-1 (or even +- a couple hundreds) blocks.

bridge-deploy/tests/test_cli.py
cducrest: I am not sure how we will deploy the reward contract. I think a regular deployment is the easiest

bridge-deploy/bridge_deploy/cli.py
saschagoebel: Done.

bridge-deploy/bridge_deploy/home.py
saschagoebel: We don't need the block number for the initialize, but for the docker-compose based validator application.

bridge-deploy/bridge_deploy/cli.py
saschagoebel: ```transaction_options``` is actually a Dict (just checked) and most of the parameters are optional.
Personally I'd leave it as is for now, as all init params are required and serve a specific purpose.
Also, with the removal of the foreign limits & owner, the list isn't that long anymore ;-)
Fine?

cducrest: fine.

bridge-deploy/bridge_deploy/cli.py
```diff
@@ -0,0 +1,263 @@
...
+    type=str,
+    required=True,
+    callback=validate_address,
+    metavar="BLOCK_REWARD_ADDRESS",
+    envvar="BLOCK_REWARD_ADDRESS",
+)
+
+home_daily_limit_option = click.option(
+    "--home-daily-limit",
+    help="The daily transfer limit for the home bridge in WEI",
```
jannikluhn: can we convert this to ETH? Too easy to get the digits wrong if there are so many of them.

bridge-deploy/tests/test_cli.py
```diff
@@ -0,0 +1,54 @@
...
+@pytest.fixture
+def runner():
+    return CliRunner()
+
+
+def test_deploy_reward(runner):
+
+    result = runner.invoke(main, args="deploy-reward --jsonrpc test")
+
+    assert result.exit_code == 0
```
jannikluhn: shouldn't we also test here and in related functions that the contracts have been successfully deployed? E.g. by checking that there's code at the address? Or do we do this in other tests already?

bridge-deploy/tests/test_cli.py
saschagoebel: Reward Contract will be deployed as part of the bridge deployer now.

bridge-deploy/tests/test_cli.py
saschagoebel: That's done in the test_deploy.py.

test_cli.py just check's that the CLI interface does "something" ;-)

bridge-deploy/bridge_deploy/cli.py
saschagoebel: I aggree with the difficulty, but if we go with ETH, we probably have to handle decimals. Is there a "fool proof" way to do this with click? @cducrest @schmir-at-bb

jannikluhn: probably along these lines:

```Py
from decimal import Decimal, InvalidOperation

def validate_eth_amount(ctx, param, value):
    try:
        eth_decimal = Decimal(value)
    except InvalidOperation:
        raise click.BadParameter(f"{value} is not a valid decimal number")
    else:
        wei_amount = eth_decimal * WEI_PER_ETH

    if int(wei_amount) != wei_amount:
        raise click.BadParameter(f"{value} contains fractional Wei component")

    return int(wei_amount)

@click.option(..., type=str, callback=validate_eth_amount, ...)
...
```

saschagoebel: Nice! Worked almost out of the box. There's still an issue with the Decimal conversion. Added a TODO.

saschagoebel: Updated.

PR #157, opened 2019-06-07
Use 'petersburg' as evm target version

schmir-at-bb:
I didn't touch the bridge, but everything else should now use petersburg.


schmir-at-bb:
One note: Some of the pytest.ini files are not necessary, because the tests use code compiled by the makefile. I'd still like to add them in someone starts compiling code from the tests.

Issue #105, opened 2019-05-23
Deploy script for bridge contracts on main chain


berndbohmeier:

weilbith:
Code is so far done, using the home deployment as template.
Waiting for PR #151 to become merged, since this is based on this. I will apply all the changes due to the CR afterwards if they are necessary.

weilbith:
Done by #161
Automatic closing has not worked out...

PR #161, opened 2019-06-13
Implement foreign bridge deployment command

weilbith:
Closes to: #105

### DIFF
--- a/bridge-deploy/bridge_deploy/foreign.py
+++ b/bridge-deploy/bridge_deploy/foreign.py
@@ -1,35 +1,23 @@
-from typing import Dict, NamedTuple
+from typing import Dict
 from web3.contract import Contract
-from deploy_tools.deploy import (
-    deploy_compiled_contract,
-    increase_transaction_options_nonce,
-)
+from deploy_tools.deploy import deploy_compiled_contract
 
 from bridge_deploy.utils import load_build_contract
 
 
-class DeployedForeignBridgeResult(NamedTuple):
-    foreign_bridge: Contract
-    foreign_bridge_block_number: int
-
-
 def deploy_foreign_bridge_contract(
     *, web3, transaction_options: Dict = None, private_key=None
-):
+) -> Contract:
 
     if transaction_options is None:
         transaction_options = {}
 
-    latest_block = web3.eth.getBlock("latest")
-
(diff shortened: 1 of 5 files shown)


weilbith:
@cducrest should the `TrustlinesToken` be part of this deployment command or is it related to the merkle drop?

bridge-deploy/bridge_deploy/foreign.py
```diff
@@ -0,0 +1,35 @@
...
+    foreign_bridge_contract = deploy_compiled_contract(
+        abi=foreign_bridge_src["abi"],
+        bytecode=foreign_bridge_src["bytecode"],
+        web3=web3,
+        transaction_options=transaction_options,
+        private_key=private_key,
+    )
+    increase_transaction_options_nonce(transaction_options)
+
+    return DeployedForeignBridgeResult(foreign_bridge_contract, latest_block.number)
```
cducrest: I still don't get why we do this weird thing where we give `web3` to a function and it returns `web3.eth.getBlock("latest")` to us in a weird `DeployedForeignBridgeResult` output. Just return the contract in this function and get the latest block number in the cli.

cducrest: Otherwise the rest is fine

weilbith: I just took results of the home deployment as template and don't start to do it in another way. I though you would have discussed this at the responsive PR.
So if I do so, should I also adjust the home bridge deployment part regarding this part? @saschagoebel your comment here please as well.

cducrest: I went back to look for his comment on the other PR that I did not see before: 
"We don't need the block number for the initialize, but for the docker-compose based validator application."

cducrest:
> @cducrest should the `TrustlinesToken` be part of this deployment command or is it related to the merkle drop?

I guess it should be its own deployment script.

bridge-deploy/bridge_deploy/foreign.py
saschagoebel: Fetching the latest block could indeed be in the cli function instead of the deploy function, as it's only printed, not actually used.
Sorry, my mistake.

PR #162, opened 2019-06-14
Add checks to prevent against overflows

berndbohmeier:
Add some requires to prevent from theoretically possible overflow
errors

### DIFF
--- a/contracts/contracts/DepositLocker.sol
+++ b/contracts/contracts/DepositLocker.sol
@@ -89,8 +89,11 @@ contract DepositLocker is DepositLockerInterface, Ownable {
     function deposit(uint _valuePerDepositor) public payable isInitialised isNotDeposited onlyDepositorsProxy {
         require(numberOfDepositors>0, "no depositors");
         require(_valuePerDepositor>0, "_valuePerDepositor must be positive");
-        require(_valuePerDepositor < 2 ** 128, "_valuePerDepositor must be lower than 2 ** 128 to prevent overflow");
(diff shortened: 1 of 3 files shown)


contracts/contracts/DepositLocker.sol
```diff
@@ -89,8 +89,11 @@ contract DepositLocker is DepositLockerInterface, Ownable {
     function deposit(uint _valuePerDepositor) public payable isInitialised isNotDeposited onlyDepositorsProxy {
         require(numberOfDepositors>0, "no depositors");
         require(_valuePerDepositor>0, "_valuePerDepositor must be positive");
-        require(_valuePerDepositor < 2 ** 128, "_valuePerDepositor must be lower than 2 ** 128 to prevent overflow");
-        require(msg.value == numberOfDepositors * _valuePerDepositor, "the deposit does not match the required value");
+
+        uint depositAmount = numberOfDepositors * _valuePerDepositor;
+        require(depositAmount >= _valuePerDepositor, "Overflow in depositAmount calculation");
```
jannikluhn: I'm probably misunderstanding how overflows work or missing an additional assumption, but I'm not convinced that this catches all overflows. E.g. `numberOfDepositors = 3`, `_valuePerDepositor = 51`, overflow at `100` (for simplicity). Then I would expect `depositAmount` to be `(3 * 51) % 100 = 53 > _valuePerDepositor`. Could you clarify?

berndbohmeier: No, you are completely right, this does not make sense. This check is for addition, for multiplication you need to check differently.

Issue #175, opened 2019-06-19
Makefile to install everything in root directory


cducrest:
The makefile should create a venv in the root directory and install the contracts and deploy tools.
It should have individual targets for each tool / contracts as well.

schmir-at-bb:
merged in f7acfff9ba94127d30416f47071c825

PR #182, opened 2019-06-25
Top-Level makefile, simplify circleci config 

schmir-at-bb:
This adds a top-level makefile, simplifies the circleci config and upgrades the pre-commit hooks.

see #175


contracts/Makefile
```diff
@@ -0,0 +1,33 @@
+TOP_LEVEL=$(shell cd ..; pwd)
+VIRTUAL_ENV ?= $(TOP_LEVEL)/venv
+CONTRACTS_DIRECTORY=$(TOP_LEVEL)/contracts/contracts
+
+lint: install
```
saschagoebel: Not 100% sure, but I think ```install-requirements``` is sufficient here. *test* as well.

tools/bridge-deploy/Makefile
```diff
@@ -44,4 +44,8 @@ install: install-requirements compile
 $(VIRTUAL_ENV):
 	python3 -m venv $@
 
-.PHONY: install install-requirements test lint compile build format
+clean:
```
saschagoebel: Remove *build* and *node_modules* from the poa submodule as well?

schmir-at-bb: I chose calling `git submodule deinit`.

PR #192, opened 2019-06-27
Print a help message when running make without arguments

schmir-at-bb:
This lists all available targets.


schmir-at-bb:
Do you have a fold executable on your fine macbook?

saschagoebel:
I actually do!

But I meant this one: https://github.com/trustlines-protocol/blockchain/pull/192/files#diff-b67911656ef5d18c4ae36cb6741b7965R9

schmir-at-bb:
Don't worry, I've copied that from stackoverflow.

PR #195, opened 2019-06-27
Add init function for block reward contract

### LINKED ISSUE #193: Add init function to block reward contract
The home bridge contract and the block reward contract must know each other bidirectional. So the home bridge comes with a `initialize()` function to provide the block reward address, among other information. But the block reward uses a fixed coded list of addresses, which are whitelisted. This is quite bad for deployment. Therefore it should get a `initialize()` function as well. So both contracts can be deployed and then initialized with the information of each others opposite address.

weilbith:
Close #193

### DIFF
--- a/contracts/contracts/reward-contract/RewardByBlock.sol
+++ b/contracts/contracts/reward-contract/RewardByBlock.sol
@@ -20,11 +20,14 @@ contract RewardByBlock is EternalStorage, IRewardByBlock {
 
     // solhint-disable const-name-snakecase
     // These values must be changed before deploy
-    uint256 public constant blockRewardAmount = 1 ether;
     uint256 public constant emissionFundsAmount = 0 ether;
     address public constant emissionFunds = 0x0000000000000000000000000000000000000000;
     uint256 public constant bridgesAllowedLength = 1;
     // solhint-enable const-name-snakecase
+
+    bool public initialized = false;
+    uint256 public blockRewardAmount;
+    address public homeBridgeAddress;
 
     event AddedReceiver(uint256 amount, address indexed receiver, address indexed bridge);
     event Rewarded(address[] receivers, uint256[] rewards);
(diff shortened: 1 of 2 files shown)


contracts/contracts/reward-contract/RewardByBlock.sol
```diff
@@ -39,10 +41,25 @@ contract RewardByBlock is EternalStorage, IRewardByBlock {
...
+      for (uint i = 0; i < _homeBridgeList.length; i++) {
+        require (_homeBridgeList[i] != address(0), "Bridge can not be the zero address!");
+      }
+
+      blockRewardAmount = _blockRewardAmount;
+      homeBridgeList = _homeBridgeList;
+      initialized = true;
+    }
+
+    function addExtraReceiver(uint257 _amount, address _receiver)
```
saschagoebel: Make that uint256 ;-)

weilbith:
I don't like that we have no tests. 😑

contracts/contracts/reward-contract/RewardByBlock.sol
weilbith: Somehow my local linting for Solidity does not work correctly anymore....

contracts/contracts/reward-contract/RewardByBlock.sol
```diff
@@ -39,10 +41,25 @@ contract RewardByBlock is EternalStorage, IRewardByBlock {
         _;
     }
 
+    function initialize(uint256 _blockRewardAmount, address[] calldata _homeBridgeList) external {
```
jannikluhn: Have you thought about just accepting a single bridge address? Would make the code slightly easier and I don't see a reason why we would need more than one.

contracts/contracts/reward-contract/RewardByBlock.sol
```diff
@@ -39,10 +41,25 @@ contract RewardByBlock is EternalStorage, IRewardByBlock {
...
+      require(!initialized, "Can not be initiated twice!");
+      require(_blockRewardAmount > 0, "Block reward amount must be positive!");
```
jannikluhn: Is there anything wrong with a block reward of zero? Don't see a reason why we'd have to prevent this.

contracts/contracts/reward-contract/RewardByBlock.sol
weilbith: I'm a bit afraid to change too much of the code without having tests securing the functionality. Having only one bridge touches many more parts of the code.  Having a list but only having one entry do the same. But I can do so. :thinking:

jannikluhn: This doesn't necessarily require more changes. Just store the single bridge address somewhere, keep `bridgesAllowedLength = 1` and return the address in the `bridgesAllowed` function. This would have the added benefit of not having to remove a public function.

weilbith: Okay, yes this would keep the changes on a minimum. The code look more than ugly, but I will go for it.

weilbith: :heavy_check_mark:

contracts/contracts/reward-contract/RewardByBlock.sol
weilbith: Removed :heavy_check_mark:

weilbith:
@jannikluhn could you just re-check if this is fine? :upside_down_face:

PR #202, opened 2019-06-28
Combine and simplify home reward and bridge deploy

### LINKED ISSUE #194: Combine and extend bridge deploy commands
The bridge deployment tool does deploy and initialize the home bridge contract in a different target than the reward contract. Due to #193 they must be combined, since their deployment addresses must be put into each others initialization functions.

weilbith:
Close #194

**Warning!!!**
At the moment experiments couldn't verify what the `msg.sender` address would be for the reward contract. By the current state the bridge proxy address is enabled on the initialization of the reward contract. Probably this must be changed to the implementation.

### DIFF
--- a/tools/bridge-deploy/bridge_deploy/cli.py
+++ b/tools/bridge-deploy/bridge_deploy/cli.py
@@ -16,6 +16,7 @@ from deploy_tools.files import validate_and_format_address, InvalidAddressExcept
 
 from bridge_deploy.home import (
     deploy_home_block_reward_contract,
+    initialize_home_block_reward_contract,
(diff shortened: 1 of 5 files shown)


weilbith:
@saschagoebel was the question about the correct contract address solved? :thinking:

saschagoebel:
Only the proxy address is whitelisted at the moment and the bridge works, so I'd say yes ;-)

weilbith:
Interesting. I like it works this way, but to be honest I haven't expected it. At some point when we are bored and eager for knowledge we can try to understand how POA is doing this delegated call.

Issue #186, opened 2019-06-26
Script to make account from secret key


cducrest:
If possible, do it with parity so it can be included in quickstart script easily.
Otherwise python script within docker image.
Otherwise write python script and provide it in the docs.

cducrest:
Done.

PR #203, opened 2019-06-28
Feature/script to recover account from key

cducrest:
related: https://github.com/trustlines-protocol/blockchain/issues/186
Need to upload the docker image to trustlines-owned repo and adapt quickstart script accordingly.


schmir-at-bb:
From the top of my head: 

- The docker image should be based on ubuntu 18.04. That's the image we use for the testchain  and the user will need to download anyway.
The resulting image size from the python image is around 980MB, from the ubuntu image it'll be around 200MB
 - `__init__.py` should be removed (at least I don't know a reason why it should be there)
- We should use a symlink for the constraints.txt file
- The file should be opened with mode "w", since we're only writing to it ("w+ opens it for reading and writing)
- The shebang in the script should be '#!/usr/bin/env python3`. The `env` executuable will search for python3 in the PATH, which will give folks a chance to run it on their own machines.

I'm going to fix these issues myself and will open a new PR.

Issue #188, opened 2019-06-26
Verify that the provided keystore can be decrypted


cducrest:
Potential solution: start a python script in a docker container
start parity and look for errors

Issue #187, opened 2019-06-26
Quickstart script can import keystore files


cducrest:
JSON file keystore version 3
User should decide to either import their file or create a new one.
Must ask for the password and put it as plain text in a file. Warn user about it.

PR #204, opened 2019-07-02
Enhance quickstart

schmir-at-bb:
This enable the quickstart script to import an existing private key or keystore file as parity account. It uses a new docker image to do that.

Implements #186, #187, #188 

Supercedes PR #203

### DIFF
--- a/tools/quickstart/Makefile
+++ b/tools/quickstart/Makefile
@@ -1,0 +1,31 @@
+TOP_LEVEL=$(shell cd ../..; pwd)
+VIRTUAL_ENV ?= $(TOP_LEVEL)/venv
+
+lint: install
+	$(VIRTUAL_ENV)/bin/flake8 quickstart setup.py
+	$(VIRTUAL_ENV)/bin/black --check quickstart setup.py
+
+test:
+	@echo "We don't have any tests for the quickstart folder."
+
+install-requirements: .installed
+
+install: install-requirements
+	$(VIRTUAL_ENV)/bin/pip install -c constraints.txt -e .
+
+.installed: constraints.txt requirements.txt $(VIRTUAL_ENV)
+	$(VIRTUAL_ENV)/bin/pip install -c constraints.txt pip wheel setuptools
+	$(VIRTUAL_ENV)/bin/pip install -c constraints.txt -r requirements.txt
+	@echo "This file controls for make if the requirements in your virtual env are up to date" > $@
+
+$(VIRTUAL_ENV):
+	python3 -m venv $@
+
+clean:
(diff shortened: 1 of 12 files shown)


tools/quickstart/Makefile
```diff
@@ -0,0 +1,31 @@
...
+
+$(VIRTUAL_ENV):
+	python3 -m venv $@
+
+clean:
+	rm -rf build .tox .mypy_cache .pytest_cache */__pycache__ *.egg-info
+	rm -f .installed
+
+build-docker-image::
+	docker build -t quickstart ../.. -f Dockerfile
```
cducrest: Do we need the `../..` because we use a symlink for the constraints?

schmir-at-bb: yes, I also had to add an additonal COPY of constraints.txt file to the Dockerfile because it's a symlink (or I could have copied the whole base directory)

tools/quickstart/quickstart/cli.py
```diff
@@ -0,0 +1,104 @@
...
+def import_keystore_file(trustline_files, keystore_input_file):
+    keyfile_dict = json.loads(open(keystore_input_file, "rb").read())
+
+    while 1:
+        password = click.prompt(text="Password", hide_input=True)
+        try:
+            account = Account.from_key(Account.decrypt(keyfile_dict, password))
+            break
+        except ValueError:
+            click.echo("The password you entered is wrong. Please try again.")
```
cducrest: That, or the user did not provide a valid keyfile.

tools/quickstart/quickstart/cli.py
```diff
@@ -0,0 +1,104 @@
...
+
+    trustline_files.store(account, password)
+    sys.exit(0)
+
+
+def read_private_key():
+    while 1:
+        privkey = click.prompt(text="Private key (hex encodded)", hide_input=True)
```
cducrest: Maybe not hide input for people like Bernd who don't copy paste their private key and will struggle to enter it (idk what is best practice)

schmir-at-bb: We could also let him enter it twice :)

I thought about not hiding it, but then I decided to keep it hidden, because I assumed everyone would copy and paste it anyway. If we don't hide it, people will copy and paste it to some github issue if they run into problems. This is something I'd like to avoid.

tools/quickstart/quickstart/cli.py
schmir-at-bb: I'm not sure if that can really happen after reading the source code, though I also can't rule it out.

 I may provide another PR for it.

PR #206, opened 2019-07-04
Use the quickstart image build from the previous commit

schmir-at-bb:
Visit https://hub.docker.com/r/trustlines/quickstart/tags for the list of available tags.

### DIFF
--- a/quickstart.sh
+++ b/quickstart.sh
@@ -6,7 +6,7 @@ set -e
 # Variables
 : "${DOCKER_IMAGE_PARITY:=trustlines/tlbc-testnet}"
 DOCKER_IMAGE_WATCHTOWER="v2tec/watchtower"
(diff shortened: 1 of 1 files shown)


quickstart.sh
```diff
@@ -6,7 +6,7 @@ set -e
 # Variables
 : "${DOCKER_IMAGE_PARITY:=trustlines/tlbc-testnet}"
 DOCKER_IMAGE_WATCHTOWER="v2tec/watchtower"
-: "${DOCKER_IMAGE_QUICKSTART:=trustlines/quickstart:master0}"
+: "${DOCKER_IMAGE_QUICKSTART:=trustlines/quickstart:master3734}"
```
saschagoebel: Can't we use "latest" here?

schmir-at-bb: only if we stay compatible with what we have now forever (and if we would upload a latest image).

PR #220, opened 2019-07-09
Add burn function to foreign bridge contract

### LINKED ISSUE #216: Foreign Bridge burn function
Make the tokens of the foreign bridge burnable.

weilbith:
Close #216

### DIFF
--- a/contracts/tests/test_foreign_bridge.py
+++ b/contracts/tests/test_foreign_bridge.py
@@ -1,0 +1,42 @@
+import pytest
+
+
+@pytest.fixture()
+def foreign_bridge_with_token_balance(
+    foreign_bridge_contract, tln_token_contract, premint_token_address
+):
+    transfer_amount = 1
+
+    tln_token_contract.functions.transfer(
+        foreign_bridge_contract.address, transfer_amount
+    ).transact({"from": premint_token_address})
+
+    assert (
+        tln_token_contract.functions.balanceOf(foreign_bridge_contract.address).call()
+        == transfer_amount
+    )
+
+    return foreign_bridge_contract
+
+
+def test_burn_token(foreign_bridge_with_token_balance, tln_token_contract):
+    assert (
+        tln_token_contract.functions.balanceOf(
+            foreign_bridge_with_token_balance.address
+        ).call()
+        > 0
+    )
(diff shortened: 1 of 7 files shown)


weilbith:
I need to extend the bridge deploy tool to take an argument for the token address.

contracts/tests/test_foreign_bridge.py
```diff
@@ -0,0 +1,42 @@
...
+
+    assert (
+        tln_token_contract.functions.balanceOf(
+            foreign_bridge_with_token_balance.address
+        ).call()
+        == 0
+    )
+
+
+def test_burn_token_with_zero_balance(foreign_bridge_with_token_balance):
```
saschagoebel: Shouldn't you check the balance here?

weilbith: I agree with you above comment, that we could avoid using the zero address.

Regarding the here mentioned balance check I decided willful against it, because this is not what should be tested. That the balance changes was tested in the case before. This makes sure the burn function can be called multiple times even if there is no balance left, without getting an assert or revert.
Would you still argue for it?

saschagoebel: I'm fine with leaving it as is, but don't you think checking for 0 would be nice to have?

weilbith: :heavy_check_mark:

weilbith:
@schmir-at-bb actually there is [a test](https://github.com/trustlines-protocol/blockchain/blob/6b33e684ce5134f48190cb911393fb115328f599/contracts/tests/test_foreign_bridge.py#L42) which is named like this. Do you think this case is not correct?

schmir-at-bb:
> @schmir-at-bb actually there is [a test](https://github.com/trustlines-protocol/blockchain/blob/6b33e684ce5134f48190cb911393fb115328f599/contracts/tests/test_foreign_bridge.py#L42) which is named like this. Do you think this case is not correct?

I didn't see this in github's diff view. Sorry for the noise.

weilbith:
Blocked by @andfletcher related to trustlines-network/project#561 But since we will need this anyway, it will be probably merged soon.

PR #226, opened 2019-07-16
Remove POA bridge

jannikluhn:
Remove everything related to the POA bridge, except for the foreign bridge contract and its deploy tool.


weilbith:
Looks good to me. 
Running `rm -r .git/modules/poa-bridge-contracts` locally should do the rest.

PR #227, opened 2019-07-16
More hooks

schmir-at-bb:
This enables shfmt as pre-commit hook. We still don't have any CI integration.

I didn't enable shellcheck, I'll probably wait for the bridge shell scripts to be removed.


schmir-at-bb:
I have this on my mind. But feel free to open an issue.

PR #234, opened 2019-07-17
Prettier solidity

schmir-at-bb:
This adds autoformatting of solidity files with prettier with the solidity plugin. It's integrated with circleci and pre-commit.

I've had to disable some of the solium rules. 

This uses a forked prettier repo, because adding the solidity plugin via additional_dependencies didn't work for me.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,6 +6,13 @@ repos:
     - id: shellcheck
       exclude: ^docker/parity_wrapper.sh$
 
+-   repo: https://github.com/trustlines-protocol/prettier
+    rev: e16e6db0fbf03e18fcf63d18c5a1479dcb6e2e75
+    hooks:
+      - id: prettier
+        entry: prettier --write --list-different
+        files: "\\.sol$"
+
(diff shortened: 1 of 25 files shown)


.pre-commit-config.yaml
```diff
@@ -26,7 +33,7 @@ repos:
     - id: debug-statements
       language_version: python3.6
     - id: end-of-file-fixer
-      exclude: ^contracts/contracts/lib/.*\.sol$
+      exclude: ^contracts/contracts/lib/.*\.sol|contracts/contracts/token/TrustlinesNetworkToken.sol$
```
berndbohmeier: why do we exclude the Token contract?

schmir-at-bb: The token contract ends with a multi-line comment. prettier formats it in a way that's not compatible with the end-of-file-fixer, because it adds an empty line at the end of the file. Moving the comment to the top of the file would have also fixed it.

see https://github.com/trustlines-protocol/blockchain/pull/234/files#diff-208f0bc21783678041037e523b43ebefR147

PR #237, opened 2019-07-18
Fetch something and put it on queue

jannikluhn:
Just to get a minimal gevent setup out and complete the user story.

### DIFF
--- a/tools/bridge/tests/conftest.py
+++ b/tools/bridge/tests/conftest.py
@@ -1,0 +1,26 @@
+import pytest
+
+from eth_tester import EthereumTester
+from web3 import Web3, EthereumTesterProvider
+
+
+@pytest.fixture
+def tester_foreign():
+    return EthereumTester()
+
+
+@pytest.fixture
+def tester_home():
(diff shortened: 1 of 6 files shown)


tools/bridge/tests/conftest.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+
+from eth_tester import EthereumTester
+from web3 import Web3, EthereumTesterProvider
+
+
+@pytest.fixture
+def tester_foreign():
```
weilbith:
```suggestion
def foreign_chain():
```

weilbith: Suggestion: I think this is more clear when used in the depths of upcoming tests. Also if fits with the naming of the `chain` fixture of the `contract-deploy-tools` plugin. Of cause the same for the home chain.

jannikluhn: As discussed, I still think `tester` is the better name as it refers to the class `EthereumTester` from the `eth-tester` library.

PR #240, opened 2019-07-18
Feature/bridge event fetcher

weilbith:
Closes trustlines-network/project#582


tools/bridge/bridge/config.py
```diff
@@ -18,15 +27,34 @@ def validate_positive_float(number: Any) -> float:
...
+    "home_rpc_url",
+    "foreign_rpc_url",
+    "token_contract_address",
+    "foreign_bridge_contract_address",
+]
 
 OPTIONAL_CONFIG_ENTRIES_WITH_DEFAULTS: Dict[str, Any] = {
-    "transfer_event_poll_interval": 5
+    "max_reorg_depth": 10,
+    "event_filter_fetch_limit": 950,
```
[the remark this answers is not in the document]
jannikluhn: Also, I don't think `filter` should be in the name as we're not registering a filter. Maybe `transfer_event_fetch_limit` (not sure about that one either though)?

tools/bridge/bridge/transfer_fetcher.py
```diff
@@ -1,25 +1,135 @@
...
+        ],
+        "name": "Transfer",
+        "type": "event",
+    }
+]
+
+TRANSFER_EVENT_SIGNATURE_HASH = Web3.keccak(text="Transfer(address,address,uint256)")
+
+
+class TransferFetcher:
```
jannikluhn: A class feels a little unnecessary here, can't we just use a function? I'm imagining something like this:

```
def fetch_token_transfers(queue, w3, max_reorg_depth, transfer_fetch_limit, poll_interval):
    transfers_fetched_until = 0
    while True:
        block_number = w3.eth.blockNumber
        fetch_until = max(0, block_number - max_reorg_depth)
        if fetch_until > transfers_fetched_until:
            transfers = fetch_transfers_in_range(contract, transfers_fetched_until + 1, fetch_until)
            for transfer in transfers:
                queue.put(transfer)
    gevent.sleep(poll_interval)

def fetch_transfers_in_range(contract, start, end, transfer_fetch_limit):
    start_block_numbers = range(start + 1, end, transfer_fetch_limit)
    event_bunches = [...]
    return list(itertools.chain(event_bunches))
```

The advantage would be that it's easier to reason about, easier to test, and less likely to be abused for other stuff in the future.

tools/bridge/bridge/transfer_fetcher.py
```diff
@@ -1,25 +1,135 @@
...
+    def __init__(
+        self,
+        *,
+        queue: Queue,
+        w3_foreign: Web3,
+        token_contract_address: str,
```
jannikluhn: Suggestion: Use the web3 contract object instead as it's provides a easy to use ABI, assuming it's possible to use `getLogs` instead of `getFilterLogs` with that object.

tools/bridge/bridge/transfer_fetcher.py
```diff
@@ -1,25 +1,135 @@
...
+            foreign_bridge_contract_address
+        )
+
+        if not foreign_bridge_contract_code:
+            raise ValueError(
+                f"The given foreign bridge contract address {foreign_bridge_contract_address} "
+                "does not point to a contract!"
+            )
+
+        self._logger = logging.getLogger("bridge.transfer_fetcher")
```
jannikluhn: I only recently learned this: The underscore prefix signals "private", not "protected". I guess almost all instance variables should rather be protected than private, so I think we should drop pretty much all of those underscores here. Not sure what the rest of the team thinks about this though.

tools/bridge/bridge/transfer_fetcher.py
```diff
@@ -1,25 +1,135 @@
...
+        self._logger.debug(
+            "Fetch transfer events from block {from_block_number} to {to_block_number}."
+        )
+
+        assert from_block_number <= to_block_number, (
+            f"Can not fetch from a negative range of blocks "
+            f"({from_block_number} to {to_block_number})"
+        )
+
+        transfer_event_filter = self._token_contract.events.Transfer.createFilter(
```
jannikluhn: We should not use `createFilter`, it won't work on infura. I *think* this is the right method: https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.getLogs (not sure if there's one on the contract object as well).

tools/bridge/tests/conftest.py
```diff
@@ -24,3 +27,50 @@ def w3_foreign(tester_foreign):
...
     provider = EthereumTesterProvider(tester_home)
     return Web3(provider)
+
+
+@pytest.fixture
+def token_transfer_event_queue():
+    return Queue()
+
+
+@pytest.fixture(scope="session")
```
jannikluhn: Any reason this has `scope="session"`?

tools/bridge/tests/conftest.py
```diff
@@ -24,3 +27,50 @@ def w3_foreign(tester_foreign):
...
+def premint_token_address(accounts):
+    return accounts[0]
+
+
+@pytest.fixture()
```
jannikluhn: `()` seem unnecessary

tools/bridge/tests/test_config.py
```diff
@@ -3,18 +3,64 @@
...
+        "home_rpc_url": "url",
+        "foreign_rpc_url": "url",
+        "token_contract_address": "0x4B0b6E093a330c00fE614B804Ad59e9b0A4FE8A9",
+        "foreign_bridge_contract_address": "0xbb1046b0fe450aa48deaff6fa474adbf972840dd",
+    }
+
+
+def test_validate_missing_keys(valid_config):
+    invalid_config = {**valid_config}
+    del invalid_config["home_rpc_url"]
```
jannikluhn: suggestion: use https://toolz.readthedocs.io/en/latest/api.html#toolz.dicttoolz.dissoc

```
invalid_config = dissoc(valid_config, "home_rpc_url")
```

tools/bridge/tests/test_config.py
```diff
@@ -3,18 +3,64 @@
...
+
+
+def test_validate_unknown_keys(valid_config):
+    with pytest.raises(ValueError):
+        validate_config({**valid_config, "unknown_key": 1})
+
+
+def test_validate_invalid_rpc_url(valid_config):
+    with pytest.raises(ValueError):
+        validate_config({**valid_config, "home_rpc_url": 1})
```
jannikluhn: This should be a test for `validate_rpc_url`, not for `validate_config`.

jannikluhn: Same for tests below.

tools/bridge/tests/test_transfer_fetcher.py
```diff
@@ -1,23 +1,235 @@
...
+    w3_foreign,
+):
+    assert token_transfer_event_queue.empty()
+
+    transfer_tokens_to_foreign_bridge()
+    transfer_fetcher._fetch_token_transfer_events_in_range(
+        0, w3_foreign.eth.blockNumber
+    )
+
+    assert token_transfer_event_queue.qsize() == 1
```
jannikluhn: please also test that it's the right event that has been fetched.

tools/bridge/tests/test_transfer_fetcher.py
```diff
@@ -1,23 +1,235 @@
...
+    for i in range(transfer_count):
+        transfer_tokens_to_foreign_bridge()
+
+    tester_foreign.mine_blocks(max_reorg_depth)
+    transfer_fetcher._fetch_token_transfer_events_not_seen()
+
+    assert token_transfer_event_queue.qsize() == transfer_count
+
+
+# TODO: test worker when async framework integrated
```
jannikluhn: leftover?

weilbith:
@schmir thanks for the hint. Just solved the problem. :+1:

tools/bridge/bridge/config.py
```diff
@@ -18,15 +27,34 @@ def validate_positive_float(number: Any) -> float:
     return float(number)
 
 
-REQUIRED_CONFIG_ENTRIES = ["home_rpc_url", "foreign_rpc_url"]
+def validate_checksum_address(address: Any) -> None:
+    if not Web3.isAddress(address):
```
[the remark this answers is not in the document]
weilbith: :heavy_check_mark:

tools/bridge/bridge/config.py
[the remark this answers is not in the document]
weilbith: I changed it to be configurable to have better unit tests. Testing with a limit of 950 fixed takes forever in test cases. Especially when you want to test the special border cases.
But I understand that letting the user configure it, could be critical. So maybe just add it as fixed number to the `main.py` module where the `TransferFetcher` gets instantiated?

Regarding the log limit I'm totally fine to take your suggestion. :+1:

tools/bridge/bridge/config.py
```diff
@@ -18,15 +27,34 @@ def validate_positive_float(number: Any) -> float:
...
+REQUIRED_CONFIG_ENTRIES = [
+    "home_rpc_url",
+    "foreign_rpc_url",
+    "token_contract_address",
+    "foreign_bridge_contract_address",
+]
 
 OPTIONAL_CONFIG_ENTRIES_WITH_DEFAULTS: Dict[str, Any] = {
-    "transfer_event_poll_interval": 5
+    "max_reorg_depth": 10,
```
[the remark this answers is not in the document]
weilbith: Stick with foreign chain here?
I don't care too much about long names. Better readable is more important. For the rest I have autocompletion and for the style we have _Black_.

tools/bridge/bridge/transfer_fetcher.py
weilbith: Yes this was the problem I think. Sry for the mistake. Took the contract to make the argument filter easy. I will try it out to make sure it use the correct RPC method. In the link docu they also state to create a new filter and remove it directly. I'll check this.
Btw: Do you know a nice utility function to extend a 20 byte address to a 32 byte string. Need it regarding the topic filters. Has not worked for me in the past for some reason.

tools/bridge/tests/conftest.py
weilbith: Nope. Just try to add it where possible to probably save some time. I remove it.

tools/bridge/tests/conftest.py
weilbith: :heavy_check_mark:

tools/bridge/tests/conftest.py
weilbith: :heavy_check_mark:

tools/bridge/tests/test_config.py
weilbith: Thanks for this nice hint. Like it much more. I should remember this library more often. :+1:

tools/bridge/tests/test_config.py
weilbith: :heavy_check_mark:

weilbith: Also added positive tests for them.

tools/bridge/tests/test_transfer_fetcher.py
weilbith: Nope. Must be still tested, right?

tools/bridge/tests/test_transfer_fetcher.py
weilbith: Good point
:heavy_check_mark:

tools/bridge/bridge/config.py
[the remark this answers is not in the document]
schmir-at-bb: I'd stick with the 'foreign_chain' prefix even though this is a user visible variable. The users in this case are the bridge operators, so they should be able to cope with that.

tools/bridge/bridge/transfer_fetcher.py
weilbith: I'm confused. I'm actually using the contract. But I don't want the main module to do more than providing me the configured address. So I don't see what to do here. Anyways I probably need to switch, since in the below comments we figured out that this will not use `getLogs`.

tools/bridge/bridge/transfer_fetcher.py
schmir-at-bb: `eth_getLogs` is the right method. Please do not create any filters. They do not survive a restart of the (parity/geth) node.

tools/bridge/bridge/config.py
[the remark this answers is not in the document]
weilbith: :heavy_check_mark:

tools/bridge/bridge/transfer_fetcher.py
weilbith: @schmir yes, this is definitely not correct. :+1: 
Do you probably know a such requested utility function by one of all these nice Ethreum related libraries?

tools/bridge/bridge/transfer_fetcher.py
weilbith: Okay, here are my thoughts why I decided to use a class:
 - reduce complexity of the individual functions to improve readability
 - having multiple functions allows for easier to handle test cases for each small "feature"
 - having a constructor function allows me to evaluate the given arguments (see the contract addresses) before a worker process becomes started

More arguments are not objective anymore. I think it is possible to rewrite all this stuff but I rather would not do it.
Your argument about abusement I do not understand. >.<

tools/bridge/bridge/config.py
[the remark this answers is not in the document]
jannikluhn: yeah, having it as a parameter in functions makes sense, but not so much in the config file.

tools/bridge/bridge/transfer_fetcher.py
jannikluhn: You can split functionality up into multiple functions, a class provides no benefit here. So I don't think the first two reasons are very convincing. In fact, my argument is that having a class actually increases complexity because it makes everything stateful.

>  Your argument about abusement I do not understand.

Not a big argument, but classes tend to be extended by additional functionality whereas functions tend to stay what they originally where supposed to do. So people might add code later that doesn't fit very well but kind of works, reducing the overall quality of the code.

But if you still think a class is better, I'm ok with it, it's just a suggestion.

tools/bridge/bridge/transfer_fetcher.py
jannikluhn: Yes, I'd get rid of the underscores. The point is that it's not only a signal for people who use it, but also for people who subclass it.

tools/bridge/tests/test_transfer_fetcher.py
jannikluhn: Then I don't get the comment. Do you mean you want to write more tests that test the class when run by gevent?

tools/bridge/bridge/transfer_fetcher.py
```diff
@@ -1,25 +1,135 @@
...
+        for event in events:
+            self._queue.put(event)
+
+    def _fetch_token_transfer_events_not_seen(self) -> None:
+        self._logger.debug("Fetch new transfer events.")
+        self._logger.debug(
+            f"Last fetched block number: {self._last_fetched_block_number}."
+        )
+
+        head_block_number = max(
```
jannikluhn: I think `head` is misleading as it usually refers to the head of the chain. Maybe `fetch_until`?

tools/bridge/tests/test_transfer_fetcher.py
weilbith: I must somehow make sure that the `fetch_transfer_events` is working. Therefore yes, I must run it with `gevent`. Do you think different?

tools/bridge/bridge/transfer_fetcher.py
```diff
@@ -1,25 +1,135 @@
...
+        self._logger = logging.getLogger("bridge.transfer_fetcher")
+        self._w3_foreign = w3_foreign
+        self._queue = queue
+        self._token_contract = w3_foreign.eth.contract(
+            address=token_contract_address, abi=MINIMAL_TOKEN_ABI
+        )
+        self._foreign_bridge_contract_address = foreign_bridge_contract_address
+        self._max_reorg_depth = max_reorg_depth
+        self._transfer_event_fetch_limit = transfer_event_fetch_limit
+        self._last_fetched_block_number = (
```
jannikluhn: can we get rid of this and just keep it as a variable in the method and pass it as a parameter? Less state, easier to understand

tools/bridge/tests/test_transfer_fetcher.py
jannikluhn: No, but why can't you do this already but have to wait until "async framework integrated"?

tools/bridge/bridge/transfer_fetcher.py
jannikluhn: If you're going to keep it: Have you considered making this a subclass of `gevent.Greenlet`? IIRC that's what you typically want to do with gevent, but @schmir might know better.

jannikluhn: Oh, and one more thing: We'll need the same for the confirmation event. Can we generalize this class in such a way that it can be used on any chain with any event?

tools/bridge/bridge/transfer_fetcher.py
weilbith: :heavy_check_mark:

tools/bridge/bridge/transfer_fetcher.py
schmir-at-bb: > Have you considered making this a subclass of `gevent.Greenlet`?

Just don't do that unless you have good reason to. The code as is would work with and without gevent.

tools/bridge/bridge/transfer_fetcher.py
schmir-at-bb: A utility function doing what? I've implemented decoding of the log entries returned by eth_getLogs in py-eth-index if that is what you need.

weilbith: I adjusted it to go for `web3.eth.getLogs`.
:heavy_check_mark:

tools/bridge/bridge/transfer_fetcher.py
weilbith: :heavy_check_mark:

tools/bridge/bridge/config.py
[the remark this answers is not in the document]
weilbith: :heavy_check_mark:

tools/bridge/bridge/transfer_fetcher.py
schmir-at-bb: > Oh, and one more thing: We'll need the same for the confirmation event. Can we generalize this class in such a way that it can be used on any chain with any event?

yes, please do this.

weilbith: The generalization is definitely a good idea! I'm not 100% sure about every detail, but it will be worth it. :upside_down_face: 
I would argue to argue to change the name of the module then to something like `event_fetcher`?

weilbith: @jannikluhn [here](https://github.com/trustlines-protocol/blockchain/blob/feature/bridge-e-token-transfer-fetcher-showcase/tools/bridge/bridge/transfer_fetcher.py) is a version using functions only. It has 25% more lines of code. I must carry all the arguments through all functions even not everyone is using everything. Pleas confirm that you really really want to have this. 
Despite from this the generalization will happen anyways.

jannikluhn: I think the functions could be simplified, but by now I'm okay with the class approach (considering that we'll probably have more complicated workers that don't fit well into a function and having a shared interface for all of those is probably a good idea).

tools/bridge/bridge/config.py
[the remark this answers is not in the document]
jannikluhn: yes, that's what I meant

tools/bridge/bridge/transfer_fetcher.py
weilbith: Okay, I've generalized the class for all kind of events. Please have another look onto this part. All tests have been applied, but still test the functionality using the token and foreign bridge contract as "real" example.

weilbith: :heavy_check_mark:

tools/bridge/bridge/transfer_fetcher.py
weilbith: So you would give it as argument and return it back? :thinking:

jannikluhn: No, I would give both start and end block number as an argument. Next time you give `end` as start, and so on. Then

weilbith: This is what I'm doing. The storage of `last_fetched_block_number` is for between the calls of `fetch_events_not_seen`. So he knows where he stopped last time. Then he is generating the `from_block_number` and `to_block_number` values for blocks which haven't been seen respecting the maximum reorg depth and also the event limits per RPC request.
Please call me if this is still unclear or if I just don't understand you. Probably the last case.

tools/bridge/tests/test_event_fetcher.py
```diff
@@ -0,0 +1,297 @@
...
+
+def test_instantiate_event_fetcher_with_negative_event_fetch_limit(
+    w3_foreign,
+    token_contract,
+    transfer_event_signature,
+    transfer_event_argument_filter,
+    transfer_event_queue,
+    foreign_chain_max_reorg_depth,
+):
+    with pytest.raises(AssertionError):
```
berndbohmeier: I feel you have a misconception about assertion. An assert in the code is an additional safety net to test for something that you believe is true anyways, hence the name assert (german: behaupten). You should not rely on it, and so also not test for it. Asserts can be removed. According to the official documentation the code of an assert is equivalent to:
```
if __debug__:
    if not expression1: raise AssertionError(expression2)
```

So either remove all the tests, if you dont need those assertions, or change them in to raising exceptions.

tools/bridge/bridge/event_fetcher.py
```diff
@@ -0,0 +1,119 @@
...
+        self.event_queue = event_queue
+        self.max_reorg_depth = max_reorg_depth
+
+        # Fetching starts always one block after the last one.
+        self.last_fetched_block_number = -1
+
+    def fetch_events_in_range(
+        self, from_block_number: int, to_block_number: int
+    ) -> None:
+        from_block_number = max(from_block_number, 0)
```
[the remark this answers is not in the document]
weilbith: True. I had dichotomous feelings about that. Thanks for stating it. Removed it for the next commit.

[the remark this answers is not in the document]
weilbith: I think raising our own exception makes no sense, if web3 is already doing so.

tools/bridge/bridge/event_fetcher.py
```diff
@@ -0,0 +1,119 @@
...
+        to_block_number = min(to_block_number, self.web3.eth.blockNumber)
+
+        self.logger.debug(
+            f"Fetch events from block {from_block_number} to {to_block_number}."
+        )
+
+        assert from_block_number <= to_block_number, (
```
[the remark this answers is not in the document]
weilbith: See [above conversation](https://github.com/trustlines-protocol/blockchain/pull/240#discussion_r305726764)

schmir-at-bb:
1. Please decouple fetching of the events from putting them onto the queue. If you want to put them onto the queue you should do that in some outer function. I do think we could even get rid of putting them onto queue here and let the caller do that. (i.e. you could have a method fetch_some_events that would give you new events, or an empty list of events if there aren't any new events. In that way you could even make the caller responsible for calling sleep and additionally we could signal a 'we are completely synced with chain' state). 

2. If the `for from_block_number in range(...` loop has finished, we will call gevent.sleep() unconditionally. 
But the chain may already have moved forward and we should rather fetch more events.  Please fix this.

tools/bridge/tests/test_event_fetcher.py
schmir-at-bb: I fail to see the reasoning here. If you want to make sure a certain part of the code raises an assertion, why should you not test for it?

berndbohmeier: For me testing for raising an assertion is showing a wrong assumption about assertions in general. Assertion should never get raised in right code, you should not be able to trigger them in right code, assertion could be removed. So making sure that a certain part of the code raises an assertion does not make sense.

tools/bridge/bridge/event_fetcher.py
```diff
@@ -0,0 +1,119 @@
...
+                "address": self.contract_address,
+                "topics": [self.event_signature_hash.hex()]
+                + self.event_argument_filter,
+            }
+        )
+
+        self.logger.debug(f"Found {len(events)} events.")
+
+        for event in events:
+            self.event_queue.put(event)
```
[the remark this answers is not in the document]
weilbith: I think that has been resolved by using the web3 builtin contract evetn getLogs method. Please confirm.

tools/bridge/bridge/event_fetcher.py
```diff
@@ -0,0 +1,119 @@
...
+        self.logger.debug(
+            f"Fetch events from block {from_block_number} to {to_block_number}."
+        )
+
+        assert from_block_number <= to_block_number, (
+            f"Can not fetch from a negative range of blocks "
+            f"({from_block_number} to {to_block_number})"
+        )
+
+        events = self.web3.eth.getLogs(
```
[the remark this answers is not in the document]
weilbith: We had a call for this topic. Has been worked on and implemented for the next commit.

tools/bridge/bridge/main.py
```diff
@@ -40,14 +40,24 @@ def main(config_path: str) -> None:
...
 
     transfer_event_queue = Queue()
+
+    transfer_event_fetcher = EventFetcher(
+        web3=w3_foreign,
+        contract_address=config["token_contract_address"],
+        event_signature="Transfer(address,address,uint256)",
+        event_argument_filter=[
+            None,
+            f"0x000000000000000000000000{config['foreign_bridge_contract_address'][2:]}",
```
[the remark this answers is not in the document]
weilbith: With the adjustment of using the web3 contract event related to [this conversation](https://github.com/trustlines-protocol/blockchain/pull/240#discussion_r305766263), this has been replaced with a clear readable. dictionary.

tools/bridge/bridge/event_fetcher.py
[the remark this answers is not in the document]
weilbith: Question: Is this still relevant after Ralf his comment? If so I will change it to `if.. raise`.

tools/bridge/tests/test_event_fetcher.py
weilbith: Okay so my take-away is to go for `if ... raise` blocks instead of using assertion statements.

tools/bridge/bridge/event_fetcher.py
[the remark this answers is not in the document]
weilbith: See this [conversation](https://github.com/trustlines-protocol/blockchain/pull/240#discussion_r305770531) with extended comments to this topic (include testing).

tools/bridge/bridge/event_fetcher.py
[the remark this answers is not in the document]
weilbith: Replaced with `if.. raise` code blocks within the next commit.

tools/bridge/bridge/event_fetcher.py
```diff
@@ -0,0 +1,119 @@
+import logging
+import gevent
```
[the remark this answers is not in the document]
weilbith: Thanks for this hint! :+1:

tools/bridge/bridge/event_fetcher.py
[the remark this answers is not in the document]

tools/bridge/bridge/event_fetcher.py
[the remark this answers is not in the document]

tools/bridge/bridge/event_fetcher.py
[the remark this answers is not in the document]

tools/bridge/bridge/event_fetcher.py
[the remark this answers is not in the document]

tools/bridge/tests/test_event_fetcher.py
weilbith: :heavy_check_mark:

tools/bridge/bridge/event_fetcher.py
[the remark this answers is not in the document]
berndbohmeier: Should be fine yes. We could consider using and own format, but that can be done later.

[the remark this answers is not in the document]
weilbith: Okay, great. Therefore I close this conversation.

tools/bridge/bridge/event_fetcher.py
```diff
@@ -0,0 +1,119 @@
...
+class EventFetcher:
+    def __init__(
+        self,
+        *,
+        web3: Web3,
+        contract_address: str,
+        event_signature: str,
+        event_argument_filter: List,
+        event_fetch_limit: int = 950,
+        event_queue: Queue,
```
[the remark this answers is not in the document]
weilbith: Because of the lack for a fitting type and a custom abstract class would a overshoot, I left it untyped for now. Thereby it does not restrict the interface like it does now.

weilbith:
We decided that @schmir-at-bb  will implement his outpointed weaknesses himself in a new upcoming PR. This will not change the interface how to use it and therefore upcoming PRs build on top are not bothered by this.

tools/bridge/tests/test_event_fetcher.py
```diff
@@ -0,0 +1,310 @@
...
+
+    assert transfer_event_queue.qsize() == 1
+
+    event = transfer_event_queue.get()
+
+    assert event["event"] == transfer_event_name
+
+    for argument_name, argument_value in transfer_event_argument_filter.items():
+        print(f"Name: {argument_name} Value: {argument_value}")
+        event.args[argument_name] == argument_value
```
berndbohmeier: Is there an assert missing? At the moment this is a NOP?

weilbith: Yes. Inserted it while removing the obsolete print statements. Haven't observe ed it while scrolling to my changes.

weilbith: :heavy_check_mark:

PR #246, opened 2019-07-22
Feature/bridge docker compose

saschagoebel:
Closes https://github.com/trustlines-network/project/issues/584


tools/bridge/docker/docker-compose-nodes-development.yml
```diff
@@ -0,0 +1,41 @@
...
+      - "--parity-args"
+      - "--light"
+      - "--no-hardware-wallets"
+      - "--jsonrpc-apis=safe"
+      - "--jsonrpc-hosts=all"
+      - "--jsonrpc-cors=all"
+      - "--no-ipc"
+      - "--no-secretstore"
+    volumes: ['./data/laika:/home/parity/.local/share/io.parity.ethereum/chains_light']
+    restart: unless-stopped
```
weilbith: Suggestion: Possibly you want to use `yaml` anchor to avoid repeating of the common variables and overwrite only the few differences? :thinking:

tools/bridge/docker/docker-compose-nodes-development.yml
```diff
@@ -0,0 +1,41 @@
...
+      - "--jsonrpc-hosts=all"
+      - "--jsonrpc-cors=all"
+      - "--no-ipc"
+      - "--no-secretstore"
+      - "--no-color"
+    volumes: ['./data/ropsten:/data']
+    restart: unless-stopped
+    mem_limit: 2G
+    mem_reservation: 16M
+    stop_grace_period: 1m
```
weilbith: Question: I think Parity's own default is 3min. Do we want to quit earlier then there are problems/

tools/bridge/docker/docker-compose.yml
```diff
@@ -0,0 +1,11 @@
+version: '2.4'
+services:
+
+  validator:
```
weilbith:
```suggestion
  bridge_validator:
```

weilbith: Strong suggestion :wink:

tools/bridge/docker/docker-compose-nodes-development.yml
saschagoebel: Changed.

tools/bridge/docker/docker-compose.yml
saschagoebel: The suggested project name is tlbc-bridge, so the container name will be tlbc-bridge_validator_1

tools/bridge/docker/docker-compose-nodes-development.yml
saschagoebel: I'd rather have it more explicit and readable than short unless there's a need to shorten it.

saschagoebel:
Having the compose files separated allows to easily just start the light nodes for example. It's not technically fancy, but easy to use and grasp.

PR #247, opened 2019-07-22
Purge confirmations from ex validators

schmir-at-bb:
This makes sure that only active confirmations from active validators are taken into account.

At the moment this doesn't handle the case, where the number of confirmations would already suffice when the validator set shrinks.

### DIFF
--- a/contracts/contracts/bridge/HomeBridge.sol
+++ b/contracts/contracts/bridge/HomeBridge.sol
@@ -90,18 +90,35 @@ contract HomeBridge {
         }
     }
 
-    function _checkSufficientNumberOfConfirmations(uint16 numConfirmations)
+    function purgeConfirmationsFromExValidators(bytes32 transferStateId)
         internal
-        view
-        returns (bool)
     {
-        uint numRequired = (
-                validatorProxy.numberOfValidators() *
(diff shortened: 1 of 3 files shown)


contracts/contracts/bridge/HomeBridge.sol
```diff
@@ -90,18 +90,35 @@ contract HomeBridge {
...
-        return numConfirmations >= numRequired;
+        address payable[] storage confirmingValidators = transferState[transferStateId]
+            .confirmingValidators;
+
+        uint i = 0;
+        while (i < confirmingValidators.length) {
+            if (validatorProxy.isValidator(confirmingValidators[i])) {
+                i++;
+            } else {
+                confirmingValidators[i] = confirmingValidators[confirmingValidators
```
saschagoebel: I hope this is because of gas/storage optimization, because it's definitely not straightforward ;-)

schmir-at-bb: Do you mean deleting elements from the array?

saschagoebel: Yes. I see what the code does and it seems to work, it's just not that obvious. But rewriting the array is probably pretty costly. So 👍

PR #249, opened 2019-07-22
Implement confirmation sender

jannikluhn:
Send a confirmation transaction for every transfer event it reads on a queue. We keep track of pending transactions until they've been confirmed `max_reorg_depth` times so that we can resend them with a higher gas price if necessary (not implemented). For now we use the set of pending transactions to determine the nonce. In the future we probably want to switch to inspecting the tx pool in Parity instead, but this is easier to implement for now.


schmir-at-bb:
Error handling is missing, we need to take care of that later.

There is a possible race condition in `clear_confirmed_transactions`. The current block number is being fetched after the transaction receipts. This is being used to determine if the transactions are final.
The current block number could have moved on by `max_reorg_depth` blocks after the receipts had been fetched. To fix this, the current block number must be fetched before the receipts.

Also, `clear_confirmed_transaction` is iterating over `self.pending_transactions`, which is being modified by the other greenlet. This may not be a bug, but it's at least not nice.

The code is iterating over all transactions, though it could stop when it sees a transaction after the confirmation_threshhold. It makes the code more complex than it needs to be.

Please change the `pending_transactions` field to be Queue and handle the pending transactions one by one. This will clarify what is really happening and has the added benefit that we could use a bounded Queue.

`constants.STEP_INTERVAL` is a terrible name. Please either move it to the confirmation_sender module or give it a more descriptive name.

Suggestion: Do not iterate over `transfer_queue` in `send_confirmation_transactions` with a for loop. Queues being iterable is only true for the gevent queues, not for the stdlib queues. Having a `transfer_queue.get()` is most probably more familiar to the reader.

(I didn't look at the tests until now and will defer that to later)

jannikluhn:
> Please change the pending_transactions field to be Queue

Done, I like the change. Only drawback is that it makes it harder to pick nonces based on the pending transactions, but since we should do that based on the node's transaction pool anyways I just removed that functionality altogether.

> constants.STEP_INTERVAL is a terrible name. Please either move it to the confirmation_sender module or give it a more descriptive name.

I'm actually quite happy with the name. It refers to the step interval of the trustlines chain. Maybe `HOME_CHAIN_STEP_INTERVAL` would make this clearer, or do you have a different suggestion?

> Suggestion: Do not iterate over transfer_queue in send_confirmation_transactions with a for loop. Queues being iterable is only true for the gevent queues, not for the stdlib queues. Having a transfer_queue.get() is most probably more familiar to the reader.

To me it's actually surprising that you *can't* iterate over a Python queue. And since the for loop makes it clearer what's actually going on compared to `while True: queue.get()` I'd rather keep it like it is.

schmir-at-bb:
> Maybe HOME_CHAIN_STEP_INTERVAL

This one is fine. I didn't actually get that it refers to the value configured for our chain until now!

berndbohmeier:
> To me it's actually surprising that you can't iterate over a Python queue. And since the for loop makes it clearer what's actually going on compared to while True: queue.get() I'd rather keep it like it is.

Suggestion: Please use `get`. I agree with Ralf that this is the familiar operation on a queue.

tools/bridge/bridge/confirmation_sender.py
```diff
@@ -0,0 +1,105 @@
...
+from bridge.constants import HOME_CHAIN_STEP_INTERVAL
+
+
+class ConfirmationSender:
+    """Sends confirmTransfer transactions to the home bridge contract."""
+
+    def __init__(
+        self,
+        transfer_queue: Queue,
+        home_bridge_contract: Contract,
```
weilbith: Suggestion: I would rather not have the caller to take care about creating the contract instance. It is just complexity we don't need there. This makes even more sense if would like to continue to integrate some checks for this contract like we do for the event handler. So make sure that any contract is deployed at this address and at least make sure that the `confirmTransfer` function is there. We wanted to be better at such parts than the POA bridge. And I think this is super easy to implement, but saves a lot of possible trouble on getting a bridge validator running.
Anyways I would argue that we should try to stick with one pattern. Either the components/worker handle the blockchain components they connect to on their own or none of of them does.

tools/bridge/bridge/confirmation_sender.py
```diff
@@ -0,0 +1,105 @@
...
+        private_key: bytes,
+        gas_price: int,
+        max_reorg_depth: int,
+    ):
+        self.transfer_queue = transfer_queue
+        self.home_bridge_contract = home_bridge_contract
+        self.private_key = private_key
```
weilbith: Question: So apparently we decided for one version to handle transaction signing. Did we fixed this somewhere? Please add this design decision to the architecture wiki page if not. Else I would be happy be pointed to. :upside_down_face:

tools/bridge/bridge/confirmation_sender.py
```diff
@@ -0,0 +1,105 @@
...
+from typing import Any, Dict
+
+import gevent
+from gevent.queue import Queue
+
+from web3.contract import Contract
+from eth_keys.datatypes import PrivateKey
+from eth_utils import keccak, int_to_big_endian
+
+from bridge.constants import HOME_CHAIN_STEP_INTERVAL
```
weilbith:
```suggestion
from bridge.constants import HOME_CHAIN_STEP_DURATION
```
Nitpicking: I think if this relates to the chain specification it should have the same postfix. Moreover I think step interval doesn't even make sense. At least for this context. :man_shrugging:

tools/bridge/bridge/confirmation_sender.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def __init__(
+        self,
+        transfer_queue: Queue,
+        home_bridge_contract: Contract,
+        private_key: bytes,
+        gas_price: int,
+        max_reorg_depth: int,
+    ):
+        self.transfer_queue = transfer_queue
```
weilbith: Suggestion: I start to feel the need to distinguish between a token transfer and a bridge transfer. I think it is not much work to add a prefix where it fits, but could help everyone new digging into the code. Even us it can help to don't get confused. Probably we need to add this at other placed as well.

tools/bridge/bridge/confirmation_sender.py
```diff
@@ -0,0 +1,105 @@
...
+        return keccak(
+            b"".join(
+                [bytes(transfer.transactionHash), int_to_big_endian(transfer.logIndex)]
+            )
+        )
+
+    def send_confirmation_transaction(self, transaction):
+        self.logger.info(f"Sending confirmation transaction {transaction}")
+        self.pending_transaction_queue.put(transaction)
+        self.w3.eth.sendRawTransaction(transaction.rawTransaction)
```
weilbith: Question: I'm not sure how this will in future about persistent state etc. But could imagine that `sendRawTransaction` fails (e.g. node offline). So I would rather add the transaction to the queue after that. Else we handle a pending transaction which does not exist. But maybe I'm wrong and this doesn't matter at all.

tools/bridge/bridge/confirmation_sender.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def watch_pending_transactions(self):
+        while True:
+            self.clear_confirmed_transactions()
+            gevent.sleep(HOME_CHAIN_STEP_INTERVAL)
+
+    def clear_confirmed_transactions(self):
```
weilbith: Question: I'm a bit confused. First this isn't a requirement for trustlines-network/project#583 as I see it. You didn't referenced a specific story task, but I thought this is the one you implemented. I guess this is not bad in first place, but it just blew up this whole PR. But possibly it is necessary.
Second I fail to see the purpose of this function. Apparently nobody else does, so I'm confused. What does this else then log a message? I mean a successful confirmation will be logged anyways by the related contract event. From this point of perspective this is a big overhead. What I could image is, that it is a first step into the direction of handle stuck transactions. Thanks for helping me!

tools/bridge/bridge/confirmation_sender.py
jannikluhn: I agree that checks if the contract exists or not is useful. But to me this is validation of user input and should be done as early as possible, not here.

Reasons why I prefer passing the contract object instead of creating it in the class:
- we don't have to pass both address and ABI, so we save one argument
- it keeps the class smaller
- we might want to use the same contract objects in multiple places (can't think of an example right now though, but it doesn't seem unlikely)

tools/bridge/bridge/confirmation_sender.py
jannikluhn: I don't know if it's written down somewhere. But do we need to document every single decision? It doesn't really have any long lasting effects and it sounds like unnecessary overhead.

tools/bridge/bridge/confirmation_sender.py
weilbith: > and should be done as early as possible,

I though about it as well. My first idea was to integrate such check within the configuration validation. Unfortunately we need a web3 connection therefore already we don't have. So we would need a further module to provide such functionality. I would argue that our contract ABI constants could be that minimal, that testing for the signature hash of each entry should work out. So you could say, this is the contract, please check if the attached ABI is fully supported.

tools/bridge/bridge/confirmation_sender.py
weilbith: If you think so. For me as a user (bridge validator in this case), it is quite important. Since the set of bridge validators equals the sealing entities, I would be happy to know if I just can't use my already running validator node with unlocked account. Now we need to write a guide how to handle the valdiator key with the bridge client. So for me this is not a small decision, because it cost additional work and affects the end-user.

schmir-at-bb:
Please add docstrings to the newly created test fixtures and then feel free to merge. `pytest --fixtures` will show these and at the moment it looks pretty bad.

We may get rid of some of the calls to `gevent.sleep` in order to speedup the tests, but we can do also that later.

tools/bridge/bridge/confirmation_sender.py
jannikluhn: Main motivation for the pending transaction queue is to be able to resend transactions if they aren't confirmed (because too little gas). If a transaction is confirmed we need to remove them from the queue, that's what this method is for.

It's true that it isn't part of the requirements. I originally implemented it to be able to choose the nonce based on the transactions that are pending already (but then removed it because it's a short cut anyway and it got harder to do when I switched from a list to a queue.

tools/bridge/bridge/confirmation_sender.py
jannikluhn: well, if `sendRawTransaction` fails there's even more reason to put it on the queue as that means that we'll retry later (not implemented yet).

tools/bridge/bridge/confirmation_sender.py
weilbith: Okay I missed the early part of this PR. That's a good point. Thank you for the good explanation. :+1:

tools/bridge/bridge/confirmation_sender.py
weilbith: [Here](https://github.com/trustlines-protocol/blockchain/pull/257) is a PR focusing this problem.

PR #251, opened 2019-07-23
Support for dotenv files to easily configure the docker stack

saschagoebel:
Support for dotenv files to easily configure the docker stack (and the local app)

- Hint: .env file is ignored, because it may contain secrets, but ln -snf .env.example .env

### DIFF
--- a/tools/bridge/Makefile
+++ b/tools/bridge/Makefile
@@ -35,4 +35,4 @@ clean:
 	rm -rf build .tox .mypy_cache .pytest_cache */__pycache__ *.egg-info
 	rm -f .installed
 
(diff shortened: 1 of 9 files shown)


tools/bridge/Makefile
```diff
@@ -15,6 +15,9 @@ test: install-requirements install
 build: install
 	$(VIRTUAL_ENV)/bin/python setup.py sdist
 
+start: install
```
schmir-at-bb: please add `start` to `.phony:`

saschagoebel: Done

PR #253, opened 2019-07-23
Extend bridge client config with event fetch start

weilbith:
Closes trustlines-network/project#594

**Note:**
I changed the definition of a positive integer to include zero. This works/make sense for each configuration option which use this validator so far. It was necessary for the new options which are zero even by default.


tools/bridge/bridge/event_fetcher.py
```diff
@@ -56,7 +62,7 @@ def __init__(
         self.max_reorg_depth = max_reorg_depth
 
         # Fetching starts always one block after the last one.
```
saschagoebel: The way it's implemented, it's one block BEFORE the configured one

tools/bridge/bridge/config.py
```diff
@@ -47,6 +47,8 @@ def validate_checksum_address(address: Any) -> None:
 OPTIONAL_CONFIG_ENTRIES_WITH_DEFAULTS: Dict[str, Any] = {
     "foreign_chain_max_reorg_depth": 10,
     "transfer_event_poll_interval": 5,
+    "foreign_chain_event_fetch_start_block_number": 0,
```
saschagoebel: That's a really declarative name

tools/bridge/bridge/config.py
```diff
@@ -19,8 +19,8 @@ def validate_rpc_url(url: Any) -> None:
 def validate_positive_integer(number: Any) -> float:
     if not isinstance(number, int):
         raise ValueError(f"{number} is not an integer")
-    if number <= 0:
-        raise ValueError(f"{number} must be positive")
+    if number < 0:
```
jannikluhn: that's not a positive integer anymore. Please rename the function, e.g. `validate_non_negative_integer`.

tools/bridge/tests/test_event_fetcher.py
```diff
@@ -140,6 +153,28 @@ def test_instantiate_event_fetcher_with_negative_max_reorg_depth(
             event_argument_filter=transfer_event_argument_filter,
             event_queue=transfer_event_queue,
             max_reorg_depth=-1,
+            start_block_number=foreign_chain_event_fetch_start_block_number,
+        )
+
+
+def test_instantiate_event_fetcher_with_negative_start_block_number(
```
jannikluhn: We seem to have a lot of tests like this which are very verbose. Suggestion: introduce a fixture `event_fetcher_init_kwargs` that returns (valid) key word arguments for `__init__` and then change only the relevant argument in the test. This would make the test much shorter:

```Py
with pytest.raises(ValueError):
    return EventFetcher(**assoc(event_fetcher_kwargs, "start_block_number", -1)
```

tools/bridge/bridge/event_fetcher.py
jannikluhn: Just starting at `start_block_number` should be fine, considering that there will never be any transfer events in the genesis block.

tools/bridge/bridge/config.py
jannikluhn: `transfer_event_fetch_start_block_number` would be a *little* bit shorter and more consistent with `transfer_event_poll_interval`.

weilbith: For the moment okay, but I guess we need to generalize this in future anyways. Or would you introduce such option per event? We must differ at least between both chains.

tools/bridge/bridge/config.py
weilbith: I learned that the mathematicians can't decide between both definitions, but I don't care so I apply it. :laughing:

tools/bridge/tests/test_event_fetcher.py
weilbith: Nice idea. Due to black we can save one more line per test case if we don't use `assoc`, but `**{**event_fetcher_kwargs, "start_block_number": -1}`.

weilbith: This also makes modifications for tests at the bottom with "special" cases cleaner.

tools/bridge/bridge/event_fetcher.py

tools/bridge/bridge/config.py
weilbith: :heavy_check_mark:

tools/bridge/tests/test_event_fetcher.py
weilbith: :heavy_check_mark:

tools/bridge/bridge/config.py
weilbith: Here we go. As stated I would avoid to go for event specific configuration (the polling is bad enough, overseen when it becomes obsolete). Therefore this configuration refers to the following information:
 - relates to the `foreign_`- / `home_chain` (we need to differ at least between these both)
 - this is about the `block_number`(s)
 - describes where to `start` (kinda essential for the meaning)
 - is used for the event fetching

I would argue that the usage does not matter that much. It could be also used for different purpose in future than fetching event (even I don't see so far and probably never will). Therefore `foreign_chain_start_block_number` would be an alternative without loosing "important" information. This is 10 characters less and even 6 less than @jannikluhn his suggestion. 
@saschagoebel @jannikluhn are you fine with this name? :upside_down_face:

weilbith: We can even discuss if we strike out `chain` from all config options and hope it is intuitive.

weilbith: I made a "proposal" commit. I'll rewrite it if I get feedback from you. Else please feel free to close the conversation and merge this PR.

jannikluhn: `foreign_chain_start_block_number` sounds a bit weird (start of the foreign chain?), but I don't think it matters too much. Changing it later should be trivial if we want to.

tools/bridge/bridge/event_fetcher.py
```diff
@@ -40,9 +41,14 @@ def __init__(
         if event_fetch_limit <= 0:
             raise ValueError("Can not fetch events with zero or negative limit!")
 
-        if max_reorg_depth <= 0:
+        if max_reorg_depth < 0:
             raise ValueError("Invalid maximum reorg depth with zero or negative value!")
```
schmir-at-bb: It's not zero, please fix the error message or the test. I don't even know why this has to be changed in this PR.

tools/bridge/tests/test_event_fetcher.py
```diff
@@ -271,20 +261,44 @@ def test_fetch_events_not_seen_handle_event_limit_not_exact_multiplicate(
...
+        **{**transfer_event_fetcher_init_kwargs, "max_reorg_depth": 0}
+    )
 
     try:
         greenlet = Greenlet.spawn(transfer_event_fetcher.fetch_events, poll_time)
         transfer_tokens_to_foreign_bridge()
         sleep(poll_time + 1)
 
-        assert transfer_event_queue.qsize() == 1
+        #  assert transfer_event_queue.qsize() == 1
```
schmir-at-bb: Why is this outcommented?

tools/bridge/bridge/event_fetcher.py
weilbith: It is possible to set it to zero by the configuration option. I don't see why not. But since you like to have it, I need to introduce another validation function. Afterwards I can "expand" this check again.

tools/bridge/bridge/config.py
schmir-at-bb: > I learned that the mathematicians can't decide between both definitions, but I don't care so I apply it. laughing

zero has never been a positive number. Please stop spamming github issues.

tools/bridge/tests/test_event_fetcher.py
weilbith: Because I'm silly and haven't read again my changes until the last line on too many changes. Will commented out of cause and the test case still succeed.

tools/bridge/bridge/event_fetcher.py
schmir-at-bb: The error message doesn't match the "less than zero" test. If zero is allowed here, then change the error message.

weilbith: Ah well. Good point!!!

weilbith: Message applied. :heavy_check_mark:

tools/bridge/tests/test_event_fetcher.py
weilbith: :heavy_check_mark:

tools/bridge/bridge/config.py
weilbith: Okay, since we fail to come up with a better solution and the old version is the most descriptive one without loosing information, I go for the `Changing it later should be trivial if we want to.`.

weilbith:
@schmir I added the inline documentation for all fixtures in the event fetch tests. I hope some of them are not too trivial and the user of them understands their purpose.
I agree with you, that the old configuration option names were more descriptive and we should not mind long names here. They are no command line argument names and the users will copy the example environment file anyways. :+1:

PR #257, opened 2019-07-24
Export bridge client contract validation

weilbith:
See this conversation:
https://github.com/trustlines-protocol/blockchain/pull/249/files/da7adf2233532a7f5b3eccc7b591113f9a3505fd#r306206958


tools/bridge/bridge/main.py
```diff
@@ -53,12 +54,19 @@ def main(config_path: str) -> None:
         address=config["home_bridge_contract_address"], abi=HOME_BRIDGE_ABI
     )
 
+    validate_contract(home_bridge_contract)
+
+    token_contract = w3_foreign.eth.contract(
+        address=config["token_contract_address"], abi=MINIMAL_ERC20_TOKEN_ABI
+    )
+
```
jannikluhn: nitpick: seems like an excessive line break

tools/bridge/bridge/contract_validator.py
```diff
@@ -0,0 +1,31 @@
...
+    if not contract_code:
+        raise ValueError(
+            f"The given contract address {contract.address} does not point to a contract!"
+        )
+
+    for entry in contract.abi:
+        signature = abi_to_signature(entry)
+        signature_hash = Web3.keccak(text=signature)
+
+        if signature_hash not in contract_code:
```
jannikluhn: I'm surprised that this works: As far as I know, for function calls only the first four bytes of the hash are used, so storing the whole hash seems unnecessary. For events I don't see a reason at all why the hash would appear in the code. Do you happen to know why this works or was it just guessing? Can this change with a new (minor) Solidity version?

It's a very helpful check though, so I'd keep it in any case.

weilbith: Good point. The tests where not exhaustive enough. I talked to Yoichi and he was unsure. As result [events are included fully]( https://solidity.readthedocs.io/en/develop/contracts.html?highlight=topic#low-level-interface-to-logs), while [functions only the first four bytes](https://solidity.readthedocs.io/en/v0.5.3/abi-spec.html#function-selector). So I would suggest to make a condition and for functions only check the first 4 bytes. Do you agree?

tools/bridge/bridge/main.py
weilbith: I deleted them.

weilbith: :heavy_check_mark:

tools/bridge/bridge/contract_validator.py
weilbith: I used the [Solidity ABI JSON documentation](https://solidity.readthedocs.io/en/develop/abi-spec.html#json) to make the validation stable. Moreover I switched the tests to use the home bridge contract, since we have more various description types in the internal ABI there.

weilbith: :heavy_check_mark:

tools/bridge/bridge/contract_validator.py
```diff
@@ -0,0 +1,38 @@
...
+    in the ABI it will be checked if their signature hash can be found within
+    the code. Throws an exception if the contract could not be verified. Else
+    return nothing.
+    """
+
+    contract_code = contract.web3.eth.getCode(contract.address)
+
+    if not contract_code:
+        raise ValueError(
+            f"The given contract address {contract.address} does not point to a contract!"
```
saschagoebel:
```suggestion
            f"The given contract address {contract.address.hex()} does not point to a contract!"
```

tools/bridge/bridge/event_fetcher.py
```diff
@@ -50,11 +31,11 @@ def __init__(
             )
 
         self.logger = logging.getLogger(
-            f"bridge.event_fetcher.{contract_address}.{event_signature}"
+            f"bridge.event_fetcher.{contract.address}.{event_name}"
```
saschagoebel:
```suggestion
            f"bridge.event_fetcher.{contract.address.hex()}.{event_name}"
```

tools/bridge/bridge/contract_validator.py
weilbith: :heavy_check_mark:

tools/bridge/bridge/event_fetcher.py
weilbith: Thanks for this suggestion. Good point. :laughing:

weilbith: This is of type `str` and already in hex format.

tools/bridge/bridge/contract_validator.py
weilbith: See [below](https://github.com/trustlines-protocol/blockchain/pull/257#discussion_r307758310).

tools/bridge/bridge/event_fetcher.py
weilbith: Update:
This only holds true for the module tests. The configuration module converts address options to `bytes`. I need to find a solution for both.

PR #259, opened 2019-07-25
Implement confirmation task scheduler

jannikluhn:
It's still untested, but I'm not totally confident if this approach makes sense. Therefore I've opened this PR early to request feedback on the general design first. Will add tests and clean it up afterwards.

### DIFF
--- a/tools/bridge/bridge/confirmation_task_planner.py
+++ b/tools/bridge/bridge/confirmation_task_planner.py
@@ -1,0 +1,131 @@
+import logging
+from typing import Set
+
+from gevent.event import Event
+from gevent.queue import Queue, Empty
+
+from eth_typing import Hash32
+
+from eth_utils import decode_hex, encode_hex
+
+from bridge.event_fetcher import EventFetcher
+
+
+class TransferHashRecorder:
+    def __init__(self, event_fetcher: EventFetcher) -> None:
+        self.event_fetcher = event_fetcher
+        self.seen_transfer_hashes: Set[Hash32] = set()
+
+        self.sync_status_changed = Event()
+        self.is_in_sync = False
+
+        self.logger = logging.getLogger(
+            "bridge.transfer_hash_recorder.TransferHashRecorder"
+        )
+
+    def forget_transfer_hash(self, transfer_hash: Hash32) -> None:
+        try:
+            self.seen_transfer_hashes.remove(transfer_hash)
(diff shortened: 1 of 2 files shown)


jannikluhn:
> I'm pretty sure there must be a better way to implement that.

I agree, that's why I opened the PR :)

> These events not only include the events as seen on the chain, but also 'artificial' events generated by the respective fetcher, when they see that they reached the head of the chain and wait for new blocks.
(I think you described that idea of 'artificial' events yesterday, though you didn't call it that way).

What I eventually didn't like about this is to have different types of data on the same queue (artificial and actual events). But one way to avoid this could be to not put individual events on the queue but group them by block. So an empty group would signal that (1) there were no events in the corresponding block and (2) that the corresponding block has been synced.

> The state machine would not do any IO or interact with gevent in any way. It's just data and one part of that data would be the list of confirmations to send out.

So how would this look like in practice? A class with a method `apply_event(events_in_block) -> List[Transfer]`?

(As said in the linked issue I've unassigned myself since I'm not working tomorrow in case someone else wants to give it a shot.)

schmir-at-bb:
> > I'm pretty sure there must be a better way to implement that.
> 
> I agree, that's why I opened the PR :)
> 
> > These events not only include the events as seen on the chain, but also 'artificial' events generated by the respective fetcher, when they see that they reached the head of the chain and wait for new blocks.
> > (I think you described that idea of 'artificial' events yesterday, though you didn't call it that way).
> 
> What I eventually didn't like about this is to have different types of data on the same queue (artificial and actual events). But one way to avoid this could be to not put individual events on the queue but group them by block. So an empty group would signal that (1) there were no events in the corresponding block and (2) that the corresponding block has been synced.

There's no need to group them and it's a bit simpler, though it wouldn't hurt to group them. I would try to define the exact shape and types of events that can go in using the attrs module.

> 
> > The state machine would not do any IO or interact with gevent in any way. It's just data and one part of that data would be the list of confirmations to send out.
> 
> So how would this look like in practice? A class with a method `apply_event(events_in_block) -> List[Transfer]`?

yes (or you could even store the transfers inside the instance of that class).  Then the whole thing is self-contained.
> 
> (As said in the linked issue I've unassigned myself since I'm not working tomorrow in case someone else wants to give it a shot.)

Unfortunately I didn't find enough time to work on this.

PR #268, opened 2019-07-26
Feature/bridge client account verification

### LINKED ISSUE #265: Bridge client checks if running account is a validator
To avoid confirmation transaction fails during runtime because the configured account to sign the transaction is not a validator, this should be checked early on. Thereby the bridge validator would get already a useful error message on starting the client. This avoids trouble for the user as well as on maintenance.

weilbith:
Closes #265


tools/bridge/bridge/validation_utils.py
```diff
@@ -0,0 +1,73 @@
...
+            signature_hash = signature_hash[:4]
+
+        if signature_hash not in contract_code:
+            raise ValueError(
+                f"The contract at the given address {contract.address} does not"
+                f" support the {description_type} with the following signature: {signature}!"
+            )
+
+
+def validate_confirmation_permissions(home_bridge_contract: Contract, address: str):
```
jannikluhn: I don't think we should check this as it's (I think) impossible to do properly in light of the dynamic nature of the validator set:

- A validator might not be validator right now, but wants to confirm past pending transfers
- A validator might not be validator right now, but will be soon and wants to be ready as soon as the validator set changes

tools/bridge/bridge/validation_utils.py
```diff
@@ -0,0 +1,73 @@
...
+    validator_proxy_address = home_bridge_contract.functions.validatorProxy().call()
+    validator_proxy_contract = home_bridge_contract.web3.eth.contract(
+        address=validator_proxy_address, abi=MINIMAL_VALIDATOR_PROXY_ABI
+    )
+
+    try:
+        validate_contract(validator_proxy_contract)
```
jannikluhn: This check seems unrelated to the access permissions. Maybe just remove it and assume that you got a valid contract?

tools/bridge/bridge/validation_utils.py
```diff
@@ -0,0 +1,73 @@
...
+
+    except ValueError as error:
+        raise RuntimeError(
```
jannikluhn: What's the reason to use `RuntimeError` here? It sounds somewhat scary, I think `ValueError` or a custom exception would be better.

tools/bridge/bridge/validation_utils.py
```diff
@@ -0,0 +1,73 @@
...
+            "Serious bridge setup error. The validator proxy contract at the address the home"
+            f" bridge property points to does not exist or is not intact:\n {str(error)}"
+        )
```
jannikluhn:
```suggestion
        ) from error
```

tools/bridge/bridge/validation_utils.py
```diff
@@ -0,0 +1,73 @@
...
+
+    except Exception as error:
+        raise RuntimeError(
+            "Something went wrong while trying to verify the signing accounts permission"
+            f" on the home bridge contract:\n {str(error)}"
+        )
+
+    if not permission_status:
+        raise ValueError(
+            f"The address {address} of the configured signing account has no permission"
```
jannikluhn:
```suggestion
            f"The address {to_checksum_address(address)} of the configured signing account has no permission"
```

jannikluhn:
```suggestion
            f"The address {address} is not a bridge validator"
```

conforms a little bit better to the terminology we've been using so far.

tools/bridge/tests/conftest.py
```diff
@@ -106,6 +111,32 @@ def proxy_validator_keys(proxy_validator_accounts_and_keys):
...
+
+@pytest.fixture
+def validator_address(validator_account_and_key):
+    """Address of the validator running the confirmation sender."""
+    account, _ = validator_account_and_key
+    return account
+
+
+@pytest.fixture(scope="session")
+def non_validator_address(accounts, number_of_validators):
```
jannikluhn: maybe use the same pattern as in `validator_account_and_key` to ensure that `validator_key` and `validator_address` always belong to the same account.

tools/bridge/tests/conftest.py
```diff
@@ -106,6 +111,32 @@ def proxy_validator_keys(proxy_validator_accounts_and_keys):
...
+
+
+@pytest.fixture
+def validator_address(validator_account_and_key):
+    """Address of the validator running the confirmation sender."""
+    account, _ = validator_account_and_key
+    return account
+
+
+@pytest.fixture(scope="session")
```
jannikluhn:
```suggestion
@pytest.fixture(scope="session")
```

Any reason why this should be session scoped?

tools/bridge/tests/conftest.py
```diff
@@ -106,6 +111,32 @@ def proxy_validator_keys(proxy_validator_accounts_and_keys):
...
+def non_validator_address(accounts, number_of_validators):
+    """Address of an account which is not part of the validator set"""
+    return accounts[number_of_validators]
+
+
+@pytest.fixture(scope="session")
```
jannikluhn: ditto?

tools/bridge/bridge/validation_utils.py
```diff
@@ -0,0 +1,73 @@
+from web3 import Web3
```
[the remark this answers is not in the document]
weilbith: Yes, the naming was a big problem for me here. I felt forced to change name, because I consider the account verification to not be a contract validation.

> I expected all config validation functions in here,

I understand this. The problem is that we need a two layered validation of the configuration. First is based on each option standalone and focus in first place on the type and format. I would call this a syntactically check. Moreover this layer makes sure that all necessary options are defined and default values are applied.
The second layer is for the semantic check to make sure the configured options are what they claim to be. This includes the interoperability of the configuration values. So must I use the foreign chain RPC address to verify the token contract for example. This is not possible within layer one.

That's my view to the things. Help me finding a satisfying name.

tools/bridge/bridge/validation_utils.py
```diff
@@ -0,0 +1,73 @@
...
+
+    if not contract_code:
+        raise ValueError(
+            f"The given contract address {contract.address} does not point to a contract!"
+        )
+
+    for description in contract.abi:
+        description_type = description.get("type", "function")
+
+        if description_type in ("constructor", "fallback"):
```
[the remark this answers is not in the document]
weilbith: Good idea. I like it :+1:

tools/bridge/bridge/validation_utils.py
```diff
@@ -0,0 +1,73 @@
...
+            continue
+
+        signature = abi_to_signature(description)
+        signature_hash = Web3.keccak(text=signature)
+
+        if description_type == "function":
+            signature_hash = signature_hash[:4]
```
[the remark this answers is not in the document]
weilbith: I understand your concerns. I thought a while how to make it correct. Using two variables makes sense but lead to bad readable code for me in any case. Therefore I would propose an alternative approach:

```python
        signature_hash = Web3.keccak(text=signature)

        if description_type == "function":
            description_exists_in_code = signature_hash[:4] in contract_code

        else:
            description_exists_in_code = signature_hash in contract_code

        if not description_exists_in_code:
            raise ValueError(
```

 Do you think this solve all problems?

tools/bridge/bridge/validation_utils.py
```diff
@@ -0,0 +1,73 @@
...
+
+        if signature_hash not in contract_code:
+            raise ValueError(
+                f"The contract at the given address {contract.address} does not"
+                f" support the {description_type} with the following signature: {signature}!"
```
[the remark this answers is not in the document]
weilbith: Black do not break such string automatically. I'll go for it, because it fits into the pattern, that you could just delete the second line without touching the first one. But your are correct, it is not pretty and not worth. I'll put them to the end of the previously line.

tools/bridge/bridge/validation_utils.py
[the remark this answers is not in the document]
weilbith: I added this. :+1:

tools/bridge/bridge/validation_utils.py
```diff
@@ -0,0 +1,73 @@
...
+
+        signature = abi_to_signature(description)
+        signature_hash = Web3.keccak(text=signature)
+
+        if description_type == "function":
+            signature_hash = signature_hash[:4]
+
+        if signature_hash not in contract_code:
+            raise ValueError(
+                f"The contract at the given address {contract.address} does not"
```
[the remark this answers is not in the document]
weilbith: Related to #269
I apply this also to the rest of the code.

tools/bridge/bridge/validation_utils.py
```diff
@@ -0,0 +1,73 @@
...
+            "Serious bridge setup error. The validator proxy contract at the address the home"
+            f" bridge property points to does not exist or is not intact:\n {str(error)}"
+        )
+
+    try:
+        permission_status = validator_proxy_contract.functions.isValidator(
+            address
+        ).call()
+
+    except Exception as error:
```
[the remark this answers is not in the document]
weilbith: Yes, that's what it is about. This case is meant to catch anything web3 could raise without knowing what exactly. I'll add such comment.

tools/bridge/bridge/validation_utils.py
```diff
@@ -0,0 +1,73 @@
...
+    validator_proxy_contract = home_bridge_contract.web3.eth.contract(
+        address=validator_proxy_address, abi=MINIMAL_VALIDATOR_PROXY_ABI
+    )
+
+    try:
+        validate_contract(validator_proxy_contract)
+
+    except ValueError as error:
+        raise RuntimeError(
+            "Serious bridge setup error. The validator proxy contract at the address the home"
```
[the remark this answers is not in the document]
weilbith: Really? Sure I can apply it.

tools/bridge/bridge/validation_utils.py
weilbith: > A validator might not be validator right now, but wants to confirm past pending transfers

 This does not work with the current implementation of the home bridge contract. The [require](https://github.com/trustlines-protocol/blockchain/blob/5a54d4105443fcc51eaf91fe1f9537b89f7d5253/contracts/contracts/bridge/HomeBridge.sol#L52) do a plain check for the current active set.

> A validator might not be validator right now, but will be soon and wants to be ready as soon as the validator set changes

This would lead to possibly many failing transactions until he become a validator. I think the disadvantages are higher than running the bridge client just a bit later. And possibly miss a transfer, because over 50% other validators made it to start the node earlier than you or synced faster.
Possible we could move this check somewhere else before a confirmation is made. But this seem to be quite verbose to me.

tools/bridge/bridge/validation_utils.py
weilbith: It is not. Else I would not know what to respond to the user. If my call fails I would like to give the user a meaningful message if I can. Make a call to the correct contract is something we can check. This could at least help ourselves to verify our bridge setup after deployment.

tools/bridge/bridge/validation_utils.py
weilbith: See [here](https://github.com/trustlines-protocol/blockchain/pull/268#discussion_r308100876)

tools/bridge/bridge/validation_utils.py
weilbith: :+1:

tools/bridge/tests/conftest.py
weilbith: Jannik wrote me that the `session` scope should be only used for performance optimization. Therefore I removed all of them.

tools/bridge/tests/conftest.py
weilbith: See [above](https://github.com/trustlines-protocol/blockchain/pull/268#discussion_r308180905)

tools/bridge/bridge/validation_utils.py
weilbith: Okay, in a call including Sascha we decided the following:
- have an initial check which log a warning if it fails
- check the permissions before each confirmation 

This allows the flexibility for valdiators to start a node earlier and longer than their actual active period.

tools/bridge/bridge/validation_utils.py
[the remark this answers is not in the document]
weilbith: :heavy_check_mark:

tools/bridge/bridge/validation_utils.py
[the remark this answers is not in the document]
weilbith: :heavy_check_mark:

tools/bridge/bridge/validation_utils.py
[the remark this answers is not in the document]
weilbith: :heavy_check_mark:

tools/bridge/bridge/validation_utils.py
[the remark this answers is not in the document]
weilbith: :heavy_check_mark:

tools/bridge/bridge/validation_utils.py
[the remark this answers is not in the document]
weilbith: :heavy_check_mark:

tools/bridge/bridge/validation_utils.py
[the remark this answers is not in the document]
weilbith: Are you fine with this simple comment? Not sure if this has enough meaning for you.

tools/bridge/bridge/validation_utils.py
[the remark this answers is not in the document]
weilbith: :heavy_check_mark:

tools/bridge/bridge/validation_utils.py
weilbith: :heavy_check_mark:

tools/bridge/bridge/validation_utils.py
weilbith: :heavy_check_mark:

tools/bridge/tests/conftest.py
weilbith: :heavy_check_mark:

tools/bridge/bridge/validation_utils.py
weilbith: I think about removing or exporting this due to the [decision above ](https://github.com/trustlines-protocol/blockchain/pull/268#discussion_r308181643). We don't want to check this over and over again.

tools/bridge/bridge/validation_utils.py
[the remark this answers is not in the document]
jannikluhn: As I said, we shouldn't do this. In my opinion, the only reason to ever catch a plain `Exception` is if a third party library raises an `Exception` instead of something more specific. I don't think this is the case for web3.

tools/bridge/bridge/validation_utils.py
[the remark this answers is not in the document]
jannikluhn: Some suggestions without a clear preference from my side:

- `contract_validation.py`
- `chain_validation.py`
- `rpc_validation.py`

tools/bridge/bridge/validation_utils.py
weilbith: I implemented it. Please check if out and leave another review.

tools/bridge/bridge/validation_utils.py
weilbith: :heavy_check_mark:

tools/bridge/bridge/validation_utils.py
weilbith: I exported this check. It is now done as for the other contracts in the `main` module as well. Therefore a `get_validator_proxy_contract` function has been added, to preserve the functionality of deriving this contract from the home bridge contract. I don't wanted to initialize the confirmation sender with an additional contract. Not sure if you like this. Please add a review therefore.

tools/bridge/tests/conftest.py
weilbith: :heavy_check_mark:

tools/bridge/bridge/validation_utils.py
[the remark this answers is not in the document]
weilbith: Hmm okay. I removed it.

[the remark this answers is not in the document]
weilbith: :heavy_check_mark:

weilbith:
This will require re-depoyment and funding of the home contract. Else the client fails due to the minority adjusted home bridge contract.

tools/bridge/bridge/validation_utils.py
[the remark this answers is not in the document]
weilbith: Decided for `contract_validation`. I'm not 100% satisfied, but I fail to suggest anything better.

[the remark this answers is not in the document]
weilbith: :heavy_check_mark:

tools/bridge/bridge/confirmation_sender.py
```diff
@@ -54,7 +58,15 @@ def run(self):
     def send_confirmation_transactions(self):
         while True:
             transfer_event = self.transfer_event_queue.get()
+
+            if not is_bridge_validator(self.home_bridge_contract, self.address):
```
jannikluhn: Could this check result in transfers getting stuck if lots of people start their node before they become validators? They see the transfer, ignore it (because they aren't a validator *at the moment*), and no one ever confirms it.

tools/bridge/bridge/confirmation_sender.py
```diff
@@ -23,17 +23,21 @@ def __init__(
...
         max_reorg_depth: int,
     ):
+        self.logger = logging.getLogger("bridge.confirmation_sender.ConfirmationSender")
         self.private_key = private_key
         self.address = PrivateKey(self.private_key).public_key.to_canonical_address()
 
-        validate_confirmation_permissions(home_bridge_contract, self.address)
+        if not is_bridge_validator(home_bridge_contract, self.address):
+            self.logger.warning(
+                f"The address {self.address} is not a bridge validator to confirm "
```
jannikluhn: `to_checksum_address`

tools/bridge/bridge/confirmation_sender.py
```diff
@@ -54,7 +58,15 @@ def run(self):
     def send_confirmation_transactions(self):
         while True:
             transfer_event = self.transfer_event_queue.get()
+
+            if not is_bridge_validator(self.home_bridge_contract, self.address):
+                self.logger.warning(
+                    f"Can not confirm transaction because {self.address} is not a bridge validator!"
```
jannikluhn: `to_checksum_address`

tools/bridge/bridge/contract_validation.py
```diff
@@ -0,0 +1,59 @@
+from web3 import Web3
+from web3.contract import Contract
+from web3._utils.abi import abi_to_signature
+from eth_utils import to_checksum_address
+
+from bridge.contract_abis import MINIMAL_VALIDATOR_PROXY_ABI
+
+
+def validate_contract(contract: Contract) -> None:
```
jannikluhn: maybe `validate_contract_existence` as it doesn't really validate the contract itself, just that the right contract exists at the right address.

tools/bridge/bridge/contract_validation.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def get_validator_proxy_contract(home_bridge_contract: Contract) -> Contract:
+    validator_proxy_address = home_bridge_contract.functions.validatorProxy().call()
+    return home_bridge_contract.web3.eth.contract(
+        address=validator_proxy_address, abi=MINIMAL_VALIDATOR_PROXY_ABI
+    )
+
+
+def is_bridge_validator(home_bridge_contract: Contract, address: str) -> bool:
```
jannikluhn: `address` should be `bytes`

tools/bridge/tests/test_confirmation_sender.py
```diff
@@ -194,3 +195,24 @@ def test_pending_transfers_are_cleared(
...
+):
+    try:
+        greenlet = gevent.spawn(confirmation_sender_with_non_validator_account.run)
+
+        assert (
+            confirmation_sender_with_non_validator_account.pending_transaction_queue.empty()
+        )
+
+        transfer_queue.put(transfer_event)
+        gevent.sleep(0.1)
```
jannikluhn: Just a comment: I'm not too happy about this kind of test as there might be plenty of other reasons why the queue is still empty after the `gevent.sleep` (e.g. if there's some kind of RPC request if we run the test on a live chain or if there's some other `sleep` in there for whatever reason). I don't know a better way to do it though.

tools/bridge/bridge/confirmation_sender.py
weilbith: Well this is a really good point. :scream: 
To be honest I fail to come up with an immediate solution. At least not for current setup. :thinking: 
Any suggestions from your side?

tools/bridge/bridge/contract_validation.py
weilbith: I think you suggestions is good enough. I could argue that there is even more included, but I guess a matching ABI could be also count as the "correct" existence.

tools/bridge/tests/test_confirmation_sender.py
weilbith: Yeah, I also thought about it while taking the template from you. But as you I can't come up with a better solution for now. I would add a TODO comment on it, just that it is not totally forgotten. Especially when we do some live chain tests and this case fails, the developer see this comment and get an hint.

tools/bridge/bridge/confirmation_sender.py
weilbith: :heavy_check_mark:

tools/bridge/bridge/confirmation_sender.py
weilbith: :heavy_check_mark:

tools/bridge/bridge/contract_validation.py
weilbith: :heavy_check_mark:

tools/bridge/bridge/contract_validation.py
weilbith: :heavy_check_mark:

tools/bridge/tests/test_confirmation_sender.py
weilbith: :heavy_check_mark:

tools/bridge/bridge/confirmation_sender.py
weilbith: In a meeting including @saschagoebel wie decided that this must not be solved within this PR. We do not come up with a perfect solution. Therefore I opened the issue #276. The problem is not critical, but should be solved soon.

PR #272, opened 2019-07-29
Confirmation task planner take two

jannikluhn:
Cf. #259 

Tried to incorporate @schmir-at-bb suggestions, but I don't think I understood exactly what they were and I like this even less than the first version. We should have a call on this tomorrow.

### DIFF
--- a/tools/bridge/bridge/confirmation_task_planner.py
+++ b/tools/bridge/bridge/confirmation_task_planner.py
@@ -13,7 +13,9 @@ class TransferHashRecorder:
         self.timestamped_transfer_hashes: List[TimestampedTransferHash] = []
         self.current_timestamp = 0
 
-    def record_transfer_hash(self, timestamp: int, transfer_hash: Hash32) -> None:
+    def record_transfer_hash(
+        self, timestamp: int, transfer_hash: Optional[Hash32]
+    ) -> None:
         """Record an event or record that no further event happened until the given timestamp.
 
         Events or non-events must be recorded in order.
--- a/tools/bridge/bridge/confirmation_task_planner.py
+++ b/tools/bridge/bridge/confirmation_task_planner.py
@@ -1,0 +1,135 @@
+from typing import List, Set
+
+from eth_typing import Hash32
+
+
+MAX_CLOCK_DISPARITY = 60  # maximum lock disparity between home and foreign chain
+HISTORY_TIME = 60 * 60  # forget about events that are older than this time
+
+
+class TimestampedTransferHash:
+    timestamp: int
+    transfer_hash: Hash32
+
+
+class TransferHashRecorder:
+    def __init__(self):
+        self.timestamped_transfer_hashes: List[TimestampedTransferHash] = []
+        self.current_timestamp = 0
+
+    def apply_event_batch(self, timestamp, events):
+        if timestamp < self.current_timestamp:
+            raise ValueError(f"Tried to apply events out of order")
+        self.current_timestamp = timestamp
+
+        for event in events:
+            timestamped_transfer_hash = TimestampedTransferHash(
+                timestamp, events.args.transferHash
+            )
+            self.timestamped_transfer_hashes.append(timestamped_transfer_hash)
+
+    def forget_transfer_hash(self, transfer_hash: Hash32) -> None:
+        self.timestamped_transfer_hashes = [
+            timestamped_transfer_hash
+            for timestamped_transfer_hash in self.timestamped_transfer_hashes
+            if timestamped_transfer_hash.transfer_hash != transfer_hash
+        ]
+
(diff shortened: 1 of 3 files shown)


schmir-at-bb:
I'll try to take a look at it today and we could have a call on this tomorrow.

jannikluhn:
> I still do not get why we need the all these timestamps. But I'm fine with keeping it this way, I guess this will get clear to me with the followup PR.

We need some notion of time to know
- up to which point the two chains are synchronized
- at which time we can forget about stale events
- to be able to give the home chain some time after transfer events to produce confirmation/completion events

Block numbers don't work as they are not synchronized between the two chains.

schmir-at-bb:
> * up to which point the two chains are synchronized

What does that mean?

> 
> * at which time we can forget about stale events

What are stale events? If there is a transfer we either have to confirm it, or it is already confirmed or we did already confirm it.


> 
> * to be able to give the home chain some time after transfer events to produce confirmation/completion events

Here we can use the local time on the machine from what I can tell.

jannikluhn:
> > * up to which point the two chains are synchronized
> 
> What does that mean?

We see a transfer event at time `t`. If we have to confirm it or not depends on if there's a conf/comp event some time after that. To know this, we need to be sure that the home chain has synchronized at least until `t + HOME_CHAIN_LOOKAHEAD`.

> > * at which time we can forget about stale events
> 
> What are stale events? If there is a transfer we either have to confirm it, or it is already completed or we did already confirm it.

I just mean events that have no meaning for us anymore but we still keep track of them. Most common will probably be a combination of transfer+confirmation event. We should not delete the two before observing the completion event because otherwise we'd end up with a stale completion event. Other example could be a confirmation event by a rogue validator for a transfer that doesn't exist.

> > * to be able to give the home chain some time after transfer events to produce confirmation/completion events
> 
> Here we can use the local time on the machine from what I can tell.

Maybe, I don't quite know what you have in mind here, but that information would still have to come from the event fetcher, presumably via the event queue, and I don't see how this would be any simpler.

schmir-at-bb:
> ``.



> We see a transfer event at time `t`. If we have to confirm it or not depends on if there's a conf/comp event some time after that. To know this, we need to be sure that the home chain has synchronized at least until `t + HOME_CHAIN_LOOKAHEAD`.

When we are in sync with the home chain, i.e. we have 'recorded all events, we can check if we have seen a completed event or if we have seen a confirmation event from ourselves (in case we forgot the latter). That's the point of introducing  the artificial event 'synced with chain'. I don't think the HOME_CHAIN_LOOKAHEAD is necessary.

jannikluhn:
At the very least the artificial event would need a timestamp, otherwise it could be that the artificial event is outdated (i.e. the chain was synced at some point in the past, but is not anymore). And I doubt the code would get any simpler if we had proper events without and artificial events with timestamp.

Agree that the `HOME_CHAIN_LOOKAHEAD` is not strictly necessary, but it's nice to have. It also might make future improvements easier, in particular we could introduce a random lookahead to avoid >50% confirmations.

schmir-at-bb:
> At the very least the artificial event would need a timestamp, otherwise it could be that the artificial event is outdated (i.e. the chain was synced at some point in the past, but is not anymore). 
> And I doubt the code would get any simpler if we had proper events without and artificial events with timestamp.

The code would get simpler because we could get rid of all the timestamp handling. The problem is not the additional field in the events. The problem is that we do the timestam handling in the confirmation planner. The confirmation_task_planner is full of this timestamp handling. At the moment I don't see why we need this. Let me repeat myself:

If we are synced with the home chain, we can send out confirmations for every transfer that we have seen on the foreign chain up to this point, in case we did not already sent out a confirmation and we didn't see a  completion event for the transfer.  You don't need any timestamps for that.

Either the above statement is wrong or your implementation is too complicated. In the first case, please explain to me why it is wrong, in the second case, please change the implementation and get rid of the timestamp handling.

> Agree that the HOME_CHAIN_LOOKAHEAD is not strictly necessary, but it's nice to have. It also might make future improvements easier, in particular we could introduce a random lookahead to avoid >50% confirmations.

I'm not sure what your plan is. But I'd rather keep it simple.

jannikluhn:
> If we are synced with the home chain

how do we know that we are synced without a timestamp? If it's just a "you are synced now" on the queue, this information could be outdated.

schmir-at-bb:
> > If we are synced with the home chain
> 
> how do we know that we are synced without a timestamp? If it's just a "you are synced now" on the queue, this information could be outdated.

We know we are synced with the home chain, when we have fetched till the latest block and there are no new events. We put the artificial event on the queue and can even give it a timestamp. If we fail to process the event in the next say 2s we ignore it.
The fact that we use a timestamp for this one event, doesn't mean we need all the complexity that we have now.

PR #277, opened 2019-08-01
Fix event fetcher

schmir-at-bb:
This changes the event fetching as discussed. We now have a chance to put an additional event on the queue if we're synced with the chain.

It also changes multiple minor stuff and fixes some bugs.


tools/bridge/bridge/event_fetcher.py
```diff
@@ -70,36 +70,30 @@ def fetch_events_in_range(
...
-
-        for from_block_number in range(
-            self.last_fetched_block_number + 1,
-            fetch_until_block_number + 1,
-            self.event_fetch_limit,
-        ):
+        This method returns an empty list if the caller should wait
+        for new blocks to come in.
+        """
+        while True:
```
jannikluhn: I think this loop is a little hard to understand. If I understand correctly, it is only necessary because we use the empty list to denote that the caller should wait, so we can't use it to denote that there are no events. Which is a bit confusing to me, because the latter seems like the much more appropriate use for an empty list.

Have you thought about getting rid of the loop and either returning `None` or raising some kind of exception if the chain hasn't progressed since the last call? I think this would make the code a little easier to understand.

tools/bridge/bridge/event_fetcher.py
```diff
@@ -110,5 +104,8 @@ def fetch_events(self, poll_interval: int) -> None:
         self.logger.debug("Start event fetcher.")
 
         while True:
-            self.fetch_events_not_seen()
-            sleep(poll_interval)
+            events = self.fetch_some_events()
+            for event in events:
+                self.event_queue.put(event)
+            if not events:
```
jannikluhn: If I understood correctly, one of the goals of the PR was to put the artificial events on the queue if we didn't get any real ones. But it doesn't look like this is actually happening, or am I missing something?

schmir-at-bb: No, you're not missing anything. This isn't implemented yet, but we had a bug in the old code, where it called sleep, when there were already new blocks. This has been fixed.

tools/bridge/bridge/event_fetcher.py
schmir-at-bb: > I think this loop is a little hard to understand. If I understand correctly, it is only necessary because we use the empty list to denote that the caller should wait, so we can't use it to denote that there are no events

it actually denotes that the caller should wait *and* that there are no new events. Raising an exception is bad because it's not an exceptional situation. Returning None only makes it harder to handle the return value.

schmir-at-bb: it's a bit like the read function returning an empty string at EOF. To me this is a common idiom.

jannikluhn:
Can this be rebased and merged or is there still something missing? I guess timestamps and artificial events are out of scope for this PR?

schmir-at-bb:
A positive review was missing. I can merge this later.

PR #281, opened 2019-08-02
Gevent component for confirmation task planner

jannikluhn:
This depends on the event fetcher putting artificial events and timestamps on the queue. It also depends on #272 and potentially #277. So it's definitely not ready for merge, but if someone feels like it they can do a preliminary review already.


jannikluhn:
Closing as it's probably easier to rewrite this.

PR #285, opened 2019-08-05
Remove redundant code with deploy tools dependency

weilbith:
Due to the history of the deploy tools as extraction from the auction
deploy tool, a huge bunch of code is replicated in both packages.
Therefore the auction deploy tool could be simplified by re-using the
functionality of the deploy tools.
Mark that also the common click CLI options are taken from there. This
means that the associated environment variables are named differently
(`DEPLOY_TOOLS_...`). CLI options specific to the deploy tools keep the
prefix `AUCTION_DEPLOY_`

Related to: trustlines-protocol/contract-deploy-tools#53

Note: For the usage of the CLI option environment variables, this needs a new version of the `deploy-tools` by the former mentioned PR.


weilbith:
@schmir-at-bb that was the idea. Was not clear enough in the description.

schmir-at-bb:
contract-deploy-tools 0.4.4 has been released. 

Can you please make sure that you also upgrade the package in all of the repos listed here:
https://github.com/trustlines-network/project/issues/607

(note: not all of them have contract-deploy-tools as a dependency)

weilbith:
All done. :heavy_check_mark:

PR #288, opened 2019-08-06
Don't use event timestamps

jannikluhn:
Version 3 of the confirmation task planner with changes as discussed (don't use event timestamps, except for "artificial" events, and don't try to delete old events).


tools/bridge/bridge/confirmation_task_planner.py
```diff
@@ -1,161 +1,88 @@
...
+        self.transfer_hashes: Set[Hash32] = set()
+        self.confirmation_hashes: Set[Hash32] = set()
+        self.completion_hashes: Set[Hash32] = set()
+
+        self.scheduled_hashes: Set[Hash32] = set()
+
+        self.confirmations_synced_until = 0.0
+        self.completions_synced_until = 0.0
+
+    def apply_sync_completed(self, event: str, timestamp: float) -> None:
```
weilbith: According to the following `apply_event()` function, I would go for `event_name` here. Because in the following an `event` is an actual `dict` with all the attributes and the `event_name` must be extracted before the evaluation.

```suggestion
    def apply_sync_completed(self, event_name: str, timestamp: float) -> None:
```

tools/bridge/bridge/confirmation_task_planner.py
```diff
@@ -1,161 +1,88 @@
...
+            raise ValueError(f"Got unknown event {event}")
 
-        transfer_hashes = self.transfer_event_recorder.get_transfer_hashes(
-            start_time=start_time, end_time=end_time
+    def is_in_sync(self, current_time: float) -> bool:
+        synced_until = min(
+            self.confirmations_synced_until, self.completions_synced_until
         )
-        self.transfers_processed_until = transfer_time
+        return current_time <= synced_until + self.sync_persistence_time
```
weilbith: Probably we should make sure that `self.sync_persistence_time` is not negative on initialization?

PR #290, opened 2019-08-07
Gevent component of confirmation task planner

### LINKED ISSUE #235: Confirmation worker does not send unnecessary confirmations
When triggered by a `transfer` event, the confirmation worker does not send a confirmation if the transfer was already completed or if the validator already sent a confirmation.

### LINKED ISSUE #236: Home bridge fetcher worker stores confirmation state
The home bridge fetcher worker needs to store in the internal state whether a bridge transfer is already confirmed by the validator and whether it is already complete by looking at `Confirmation` and `TransferCompleted` events

jannikluhn:
depends on #288 and the not yet opened artificial event PR

Closes #236 
Closes #235


schmir-at-bb:
There's no need to spawn multiple greenlets if we use only one queue. It get's much easier.

jannikluhn:
I've tested this on a live chain and *something* is not working, but I'm unsure if this is due to the confirmation planner. It could also be the confirmation sender's or even the home bridge contract's fault. There's a single Transfer and a single Confirmation event. They appear to have two different transfer hashes (not sure if that should be the case), so the planner schedules a new transaction. The transaction fails. I have no idea why.

jannikluhn:
I've tested this between ropsten and the tlbc testnet and it seems to work. I had to do a couple of unrelated fixes/workarounds:

- web3.py raises a `TransactionNotFound` error since the last update, which broke the confirmation sender. I now catch this exception and wait an appropriate amount of time
- the confirmation sender tries to execute the transaction to determine the gas limit. This is bound to fail as at that point it doesn't know that the transaction is sent by a validator (the sender is only defined by the signature which is added later). I fixed this by hard coding the gas limit.
- The event fetcher can't filter by validator as the event isn't indexed. I added a temporary workaround and will open an issue for this later.

tools/bridge/bridge/confirmation_task_planner.py
```diff
@@ -74,7 +82,72 @@ def pull_transfers_to_confirm(self, current_time: float) -> List[AttributeDict]:
...
+    def process_home_bridge_events(self) -> None:
+        while True:
+            event = self.home_bridge_event_queue.get()
+            if isinstance(event, FetcherReachedHeadEvent):
+                self.logger.info("Home bridge is in sync now")
+                self.recorder.apply_home_chain_synced_event(event.timestamp)
+            else:
+                self.logger.info("Received home bridge event")
+                self.recorder.apply_proper_event(event)
+            self.check_for_confirmation_tasks()
```
schmir-at-bb: Why don't we move this to the end of the first if block? To me it's rather surprising why it's being called for every processed event.

tools/bridge/bridge/constants.py
```diff
@@ -3,3 +3,5 @@
 TRANSFER_EVENT_NAME = "Transfer"
 CONFIRMATION_EVENT_NAME = "Confirmation"
 COMPLETION_EVENT_NAME = "TransferCompleted"
+
+CONFIRMATION_TRANSACTION_GAS_LIMIT = 300_000
```
schmir-at-bb: What is this value? Please add a short comment describing this value!

The last confirmation that completes a transaction currently uses 320821 gas with 123 validators, so if this value is related to that you also have to increase it a bit.

You can run `test_measure_gas_home_bridge.py` to show you the actual values.

schmir-at-bb:
> There's no need to spawn multiple greenlets if we use only one queue. It get's much easier.

I'm fine now with having two separate queues for the foreign bridge and the home bridge for the following reason:
We may want to stop fetching from one of the chains if the other chain is far behind. This will be easy to do if we use bounded queues and stop taking events from one of the queues if the other is too far behind.
Spawning multiple greenlets still is not necessary, but I would change that if we implement the above.

tools/bridge/bridge/confirmation_task_planner.py
schmir-at-bb: Looking over the code I see it's also being called from `process_transfer_events`. 

I would like to keep this as simple as possible.

Please only call this when you see the `FetcherReachedHeadEvent` on the home chain.

schmir-at-bb: i.e. do not call it from `process_transfer_events` and move this into the first if block.

schmir-at-bb:
>     * the confirmation sender tries to execute the transaction to determine the gas limit. This is bound to fail as at that point it doesn't know that the transaction is sent by a validator (the sender is only defined by the signature which is added later). I fixed this by hard coding the gas limit.

The commit message only says 'Various small fixes' from what I can tell.  We should also write that somewhere down in the code why we do have the gas limit constant in the first place. 
Also please mention the `test_measure_gas_home_bridge.py` script.

(I am still kind of annoyed that I have to defend asking for a comment /cc @berndbohmeier)

Since you failed to specify a working limit, please replace the hard-coded 400_000 gas limit in the test_measure_gas_home_bridge.py with this constant. We will notice problems with a wrong constant in that way.

jannikluhn:
I think I've addressed all comments. Let me know if I've missed anything.

> Please replace the hard-coded 400_000 gas limit in the test_measure_gas_home_bridge.py with this constant.

It seems to be hard to get this constant there without adding the bridge as a requirement to the contracts. Therefore I've opted for adding comments on both sides to indicate that the two values should be kept in sync.

schmir-at-bb:
> It seems to be hard to get this constant there without adding the bridge as a requirement to the contracts. Therefore I've opted for adding comments on both sides to indicate that the two values should be kept in sync.

ok, I didn't think about that.

tools/bridge/bridge/constants.py
```diff
@@ -3,3 +3,9 @@
 TRANSFER_EVENT_NAME = "Transfer"
 CONFIRMATION_EVENT_NAME = "Confirmation"
 COMPLETION_EVENT_NAME = "TransferCompleted"
+
+# Gas limit used for confirmation transactions. The actual gas usage can be determined with
+# test_measure_gas_home_bridge.py found in the smart contract test directory. As of commit
+# cc46ea961ece850ce28a2d62b7c484f8fa82ca3c, this is 321004, but we add a generous safety margin.
+# On changin this value, update the corresponding constant in the test script accordingly.
```
schmir-at-bb: s/changin/changing/

Issue #287, opened 2019-08-06
Introduce artificial event


jannikluhn:
The confirmation task planner needs to know if we're synced or not. One way to do this is to make the event fetcher put an "artificial event" on the queue that signals the fact that we are in sync. The artificial event needs to have a timestamp as the sync status is temporary. It needs to be possible to distinguish between normal and artificial events (e.g. via an `isinstance` check). The timestamp should probably (?) use the local clock, not the timestamp in the last block as we get a much better resolution.

weilbith:
I have something super simple, since this is not much logic (~5 loc). But @jannikluhn will have a quick meeting soon to discuss the further handling and interaction of the planner.

weilbith:
There are some problems with `gevent` on the tests. I have to discuss my solution of a monkey patch with Ralf.

schmir-at-bb:
The gevent problems will be solved by #292

PR #291, opened 2019-08-07
Add synthetic event to mark sync status of bridge

weilbith:
Closes: #287

Very open for suggestions of a better name... :see_no_evil: 

The tests fail at the moment due to `gevent` when running the whole suite. I solved it locally with a monkey patch on the very top of the `conftest`. But @schmir I guess your approach with a custom `mypy` script like for the `relay` project is better/cleaner?

@schmir @jannikluhn feel free to reject this PR if it is not necessary anymore to another design change related to the event fetcher.

### DIFF
--- a/tools/bridge/bridge/event_fetcher.py
+++ b/tools/bridge/bridge/event_fetcher.py
@@ -1,11 +1,16 @@
 import logging
-from time import sleep
+from time import sleep, time
 from typing import Any, Dict, List
 
 from eth_utils import to_checksum_address
 from web3 import Web3
 from web3.contract import Contract
 from web3.datastructures import AttributeDict
+
+
+class FetcherReachedHeadEvent:
+    def __init__(self):
+        self.timestamp = int(time())
 
 
 class EventFetcher:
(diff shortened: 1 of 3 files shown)


tools/bridge/bridge/event_fetcher.py
```diff
@@ -1,12 +1,17 @@
...
 from typing import Any, Dict, List
 
 from eth_utils import to_checksum_address
 from web3 import Web3
 from web3.contract import Contract
 
 
+class FetcherReachedHeadEvent:
+    def __init__(self):
+        self.timestamp = int(time())
```
jannikluhn: what's the reason to not use `float` here?

weilbith: I was under the false impression we need the full seconds here. I just remove that. :laughing:

schmir-at-bb:
Please do not merge this before #292 has been merged.

weilbith:
Note: I rebase locally on `master` to solve conflicts properly.

Issue #298, opened 2019-08-12
Make logging configurable


berndbohmeier:
Initialize logging from a dictionary, which we read from the TOML file:
https://docs.python.org/3/library/logging.config.html#logging.config.dictConfig

PR #309, opened 2019-08-12
Configure logging subsystem via the TOML file

schmir-at-bb:
This is using logging.config.dictConfig:
https://docs.python.org/3/library/logging.config.html#logging.config.dictConfig

See https://github.com/trustlines-protocol/blockchain/issues/298 and
https://github.com/trustlines-protocol/blockchain/issues/299

### DIFF
--- a/tools/bridge/README.md
+++ b/tools/bridge/README.md
@@ -57,7 +57,7 @@ docker-compose --project-name tlbc-bridge --file ./docker/docker-compose.yml --f
 
 ### Configure logging via the TOML configuration file
 
-You can configure logging with by setting the 'logging' key in the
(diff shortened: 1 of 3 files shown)


tools/bridge/README.md
```diff
@@ -54,3 +54,20 @@ docker-compose --project-name tlbc-bridge --file ./docker/docker-compose.yml --f
 ```
 
 **There is no production Trustlines chain yet, so this won't work**
+
+### Configure logging via the TOML configuration file
+
+You can configure logging with by setting the 'logging' key in the
```
saschagoebel:
```suggestion
You can configure logging by setting the 'logging' key in the
```

Issue #312, opened 2019-08-13
Document bridge validator client configuration and usage


weilbith:
- [ ] add documentation on how to run the bridge client
 - [ ] add documentation for the configuration (file format)

PR #313, opened 2019-08-13
Extend README documentation of bridge client

weilbith:
Related to: #312

### DIFF
--- a/tools/bridge/README.md
+++ b/tools/bridge/README.md
@@ -1,64 +1,145 @@
-Trustlines Bridge Validator
-===========================
+# Trustlines Bridge Validator
 
-Building
---------
-The docker container needs to be built from the root path, so
+This is a client for validators of the token bridge to the _Trustlines_ chain.
+The client connects two the both networks of the bridge and observes token
+transfer related events. The made transfers on the foreign blockchain get then
+confirmed on the home chain, providing the signature of the validator. A bridge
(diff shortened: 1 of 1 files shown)


weilbith:
@schmir-at-bb thanks for your input!

> >    The client can be started with make start, but must me .
>
> not sure what you mean here

Haha, I also don't know. @saschagoebel this has been changed on your commit. What do you like to say? :laughing: 

> I probably wouldn't even mention the 'make start' command to keep it simple.

So what is then the purpose of this approach? Should they source the virtual environment? Or would you remove it at all and only propose the Docker image solution? 

>  think at the moment it's best not to mention that command/installation method.

I would take that. You have already opened an issue for that and I think this can be written there. So besides all you useful input, I would just remove this section for now.

schmir-at-bb:
> So what is then the purpose of this approach? Should they source the virtual environment? Or would you remove it at all and only propose the Docker image solution?

I'm fine with having starting the program with the `tlbc-bridge` program documented (This is also what you would get if we ever upload it to PyPI) and also the docker method.

The `make run`  is just for us developers that are too lazy.

weilbith:
According to our meeting and the recent PR #316, I recovered the `pip install` setup approach. The `Makefile` installation is considered as for development only. I opened #317 for the PyPi publishing.

Issue #304, opened 2019-08-12
Make sure we do not share the same web3/contract objects in multiple greenlets


schmir-at-bb:
The `home_bridge_contract` and the `w3_home` object is shared between multiple greenlets. This may lead to problems since the web3 code may not be aware of that fact. Even it might be working now, it may break in the future.

To be on the safe side, we should not share these objects between multiple greenlets.

schmir-at-bb:
ConfirmationSender and ConfirmationTaskPlanner both spawn multiple greenlets in their run method.

PR #315, opened 2019-08-14
Split the main function

schmir-at-bb:
main was a bit too long.

This will also make it a bit easier to not share the w3 connection
objects in different greenlets (see
https://github.com/trustlines-protocol/blockchain/issues/304)

Also use a pool to manage the greenlets.

Beware: I didn't test this with a real chain, so it might break stuff.

### DIFF
--- a/tools/bridge/bridge/main.py
+++ b/tools/bridge/bridge/main.py
@@ -8,8 +8,8 @@ import os
 
 import click
 import gevent
+import gevent.pool
 from eth_keys.datatypes import PrivateKey
-from gevent import Greenlet
 from gevent.queue import Queue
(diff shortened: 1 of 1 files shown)


schmir-at-bb:
> I did not manage to do a transfer though but I'm pretty sure it has nothing to do with this PR (rather my test setup).

Now I fear I may have broken stuff. Can you test with the master branch?

(I did 'wire' the queues in the wrong way at one point I think, because the naming was a bit confusing to me)

schmir-at-bb:
It looks like the current master branch is also not working (at least the filters are broken). Therefore I'm merging this one.

Issue #314, opened 2019-08-13
bridge: Fix dependency declaration, switch to setup.cfg based configuration


schmir-at-bb:
tools/bridge/setup.py fails to declare the needed dependencies. We should fix that so that a installation with pip could have a chance to succeed.

Also, please start using setup.cfg based configuration. The tools/quickstart script folder already uses that (look at setup.py and setup.cfg there).

weilbith:
@schmir-at-bb should I get rid of the `setup.py` at all? I would love to. I guess `pyproject.toml` is one step too far for the moment.

PR #316, opened 2019-08-14
Switch to setup.cfg based configuration

weilbith:
Close: #314
See also: #313

### DIFF
--- a/tools/bridge/setup.cfg
+++ b/tools/bridge/setup.cfg
@@ -5,14 +5,14 @@ description = token bridge validator client for the Trustlines chain
 
 [options]
 install_requires =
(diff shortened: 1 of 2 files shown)


weilbith:
As discussed in the meeting I added the version constraints.
@schmir-at-bb since this is my first time doing it for this team, I would be happy if you could have a quick gaze on it.

schmir-at-bb:
It looks reasonable. Did you try to install it into a fresh venv with `pip install .` ? If that works and `tlbc-bridge --help` runs, it's ready to go.

weilbith:
Yes, I have done so. :+1:

PR #319, opened 2019-08-14
Signal handlers

### LINKED ISSUE #303: Install signal handlers in the tlbc-bridge program
We need to explicitly install signal handlers for SIGINT and SIGTERM. When tlbc-bridge is started inside a docker container, both of those signal will be blocked as it's running as pid 1 inside the container. Installing a signal handler will make the program exit immediately when the user runs docker stop, otherwise he will have to wait a few seconds until docker decides to send a SIGKILL.

jannikluhn:
Closes #303

### DIFF
--- a/tools/bridge/bridge/main.py
+++ b/tools/bridge/bridge/main.py
@@ -6,6 +6,7 @@ import logging
 import logging.config
 import os
 import signal
+from functools import partial
 
 import click
 import gevent
@@ -19,6 +20,7 @@ from bridge.config import load_config
 from bridge.confirmation_sender import ConfirmationSender
 from bridge.confirmation_task_planner import ConfirmationTaskPlanner
 from bridge.constants import (
+    APPLICATION_CLEANUP_TIMEOUT,
     COMPLETION_EVENT_NAME,
     CONFIRMATION_EVENT_NAME,
     HOME_CHAIN_STEP_DURATION,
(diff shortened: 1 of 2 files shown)


jannikluhn:
What's wrong with the current approach?

schmir-at-bb:
> What's wrong with the current approach?

It prints a stacktrace (which is not printed without the signal handlers) and the other greenlets don't have a chance to cleanup AFAICT (not that they do anything at the moment).

jannikluhn:
> the other greenlets don't have a chance to cleanup

To make sure I learn something: Don't they already? I assumed they would because we do call `pool.kill` in the `finally` block. If not: Why do we call `pool.kill` in the `finally` block?

Probably to make this more explicit in any case...

tools/bridge/bridge/main.py
```diff
@@ -212,17 +230,15 @@ def main(config_path: str) -> None:
...
-        raise KeyboardInterrupt()
-
     try:
         for coroutine_and_args in coroutines_and_args:
             pool.spawn(*coroutine_and_args)
 
-        gevent.signal(signal.SIGINT, raise_keyboard_interrupt)
-        gevent.signal(signal.SIGTERM, raise_keyboard_interrupt)
+        stop_pool = partial(stop, pool, APPLICATION_CLEANUP_TIMEOUT)
+        gevent.signal(signal.SIGINT, stop_pool)
```
schmir-at-bb: You could just replace that with 
```
for signum in (signal.SIGINT, signal.SIGTERM):
    gevent.signal(signum, stop, pool, APPLICATION_CLEANUP_TIMEOUT)
```

But I'm also fine with keeping it the way it is.

Issue #306, opened 2019-08-12
send_confirmation_transactions drops transfer events to confirm when not (yet) a validator


schmir-at-bb:
There should be a better way to handle that situation. Maybe we should wait until we become a validator.

weilbith:
@schmir-at-bb is this equivalent to #276?

jannikluhn:
Suggested approach: On startup, run the event fetchers, but not the confirmation planner and sender. In addition, run a greenlet that periodically checks if we are a validator or not. Once the check succeeds, launch confirmation planner and sender.

Does this make sense?

schmir-at-bb:
This will work. If the confirmation planner is not running, the queues will be filled with events and no one is consuming events.

I would try to start everything like it is now and keep a flag in the TransferRecorder[1], if we're already allowed to confirmTransfers. The recorder would return an empty list in the pull_transfers_to_confirm() if we're not yet allowed to confirm transfers. Nevertheless the state in the recorder will be able to catch up with the current state of the chain. We should make sure that completed transfers don't stay in the recorder in order to not waste memory (even when we are not a validator yet).


Contrary to what I said in the meeting, I would not put this onto a queue, since I think it's special enough that we could also just call a method that does set this flag.

So, please start a greenlet that waits for the validator status to become true, sets this flag, and then still watches the validator status. If we stop being a validator, it's game over and we could probably abort the program.

[1] One could argue that we should use another name for the TransferRecorder then. But I'm fine with living with a bit badly named stuff.

PR #321, opened 2019-08-15
Validator Status Watcher

### LINKED ISSUE #276: Solve validator set transition problem of stuck transitions
Due to PR #268 the configured account of the bridge validator becomes verified. This means that before a confirmation transaction is send, it will be checked if this account is authorized to do so. If not the transfer event is dropped. The purpose of this is to allow bridge validators to start their client already earlier than they become active and also let the client running even becoming removed from the validator set. Even if the user gets informed about this situation by warnings from the very beginning (already before the first transfer event), transfers could get ignored, because to many validators ignore the event. This could happen during the period of switching validators, when the old ones stop and the new one are running already earlier as described before. 
A simple solution for the moment would be to convince enough valdiators to restart their bridge clients. Since the transfers are not finalized and haven't been confirmed by the account it should become handled now. This does not workout anymore, when we implement the feature that bridge clients remember their last state during restarts.

jannikluhn:
Opening PR to see CI output as tests don't run locally on my machine anymore (due to missing solidity version) and to give others a chance to adopt this PR as I'm not gonna work today.

The way I quit when we exit the validator set looks a bit weird to me, but I didn't come up with something better. Would have been much nicer with Trio!

I also don't like how the `is_validating` status is passed through the confirmation task planner to the transfer recorder. But I don't see a way to avoid this.

Closes #276 and #306

### DIFF
--- a/tools/bridge/bridge/confirmation_task_planner.py
+++ b/tools/bridge/bridge/confirmation_task_planner.py
@@ -17,6 +17,7 @@ class ConfirmationTaskPlanner:
         transfer_event_queue: Queue,
         home_bridge_event_queue: Queue,
         confirmation_task_queue: Queue,
+        is_validating: bool = True,
     ) -> None:
 
         self.recorder = TransferRecorder()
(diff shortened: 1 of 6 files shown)


jannikluhn:
> You have to start two additional greenlets just to set a flag on the recorder and exit the program. This is too complicated.

What's complicated about it? The greenlets themselves are pretty trivial and, conceptually, having some sort of explicit task "wait until we're a validator and then start" makes it easier to understand what's going on.

> Please simplify this by giving the a reference to the recorder and a shutdown function to the validator watcher. Get rid of the events and the two additonal greenlets.

I rejected this approach because it couples the validator status watcher with the confirmation planner even though they don't have anything to do with each other. Also, I tried to avoid using callbacks because they make the code hard to follow.

PR #322, opened 2019-08-15
Misc production fixes

schmir-at-bb:
- Enhance the logging output
- Do not use the eth addresses as part of the loggers name
- Update contract-deploy-tools to 0.5.0
- Log the bridge validators address when starting
- Show how to silence web3 with the log configuration
- Add FakeNetworkToken contract
- Rely on web3 filtering and remove the validator filtering hack
- Fix HOME_BRIDGE_ABI: Confirmation.validator is indexed now
- Log transaction failures in the confirmation sender

### DIFF
--- a/contracts/contracts/token/FakeNetworkToken.sol
+++ b/contracts/contracts/token/FakeNetworkToken.sol
@@ -1,0 +1,29 @@
+/*
+This contract is only used for testing. It emits Transfer events just
+like the TrustlinesNetworkToken contract.
+
+It can be deployed with the following command
+
+    deploy-tools deploy FakeNetworkToken
+
+Please note that all of the functions are public.
+*/
+
+pragma solidity ^0.5.8;
+
+contract FakeNetworkToken {
+    event Transfer(address indexed from, address indexed to, uint256 value);
+
+    function transfer(address from, address to, uint256 value) public {
+        emit Transfer(from, to, value);
+    }
+
+    /* emit multiple events in a single transaction */
+    function transfer4(address from, address to, uint256 value) public {
+        emit Transfer(from, to, value);
+        emit Transfer(from, to, value);
+
+        emit Transfer(from, to, value + 1);
+        emit Transfer(from, to, value + 2);
+    }
+}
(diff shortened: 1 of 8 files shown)


tools/bridge/bridge/confirmation_sender.py
```diff
@@ -16,6 +16,8 @@
 from bridge.event_fetcher import FetcherReachedHeadEvent
 from bridge.utils import compute_transfer_hash
 
+logger = logging.getLogger(__name__)
```
jannikluhn: As I said, I'm ok with the change, but curious to hear your arguments for having the logger on the module instead of the instance.

tools/bridge/bridge/confirmation_task_planner.py
```diff
@@ -29,43 +28,46 @@ def __init__(
...
+        logger.debug("Starting")
         try:
             greenlets = [
                 gevent.spawn(self.process_transfer_events),
                 gevent.spawn(self.process_home_bridge_events),
             ]
             gevent.joinall(greenlets, raise_error=True)
         finally:
-            self.logger.info("Stopping")
+            logger.info("Stopping")
```
jannikluhn:
```suggestion
            logger.debug("Stopping")
```

to use the same level as `Starting`

contracts/contracts/token/FakeNetworkToken.sol
```diff
@@ -0,0 +1,29 @@
+/*
```
jannikluhn: Is this contract used anywhere? I don't see anything

tools/bridge/bridge/confirmation_sender.py
schmir-at-bb: Well, I would say that the `logger = logging.getLogger(__name__)` is pretty much standard usage. You get the logger hierarchy for free, because it mirrors the python module hierarchy. There is no good reason to set it inside a class. When there is good reason, I'm fine with having it in the class (we still have it somewhere).

contracts/contracts/token/FakeNetworkToken.sol
schmir-at-bb: Not yet, I've used it for testing.

Issue #320, opened 2019-08-14
Retry web3 calls if they fail


schmir-at-bb:
We already have quite a good separation of things doing IO and things doing our 'business logic'. 
If one of the greenlets fails to do a web3 call, it will die and will take down the whole bridge.

We should catch errors from web3 and make sure the state of each greenlet doing web3 stuff stays valid.

See https://github.com/trustlines-protocol/blockchain/issues/300

PR #327, opened 2019-08-19
Use tenacity to retry event_fetcher's web3 calls

schmir-at-bb:
We retry and log warnings forever at the moment. This most probably
will need some tuning.

For documentation on tenacity please visit
https://tenacity.readthedocs.io/en/latest/

See https://github.com/trustlines-protocol/blockchain/issues/320

### DIFF
--- a/tools/bridge/bridge/event_fetcher.py
+++ b/tools/bridge/bridge/event_fetcher.py
@@ -2,6 +2,7 @@ import logging
 from time import sleep, time
 from typing import Any, Dict, List
 
+import tenacity
 from web3 import Web3
 from web3.contract import Contract
 from web3.datastructures import AttributeDict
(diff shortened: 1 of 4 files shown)


tools/bridge/bridge/event_fetcher.py
```diff
@@ -48,14 +49,42 @@ def __init__(
         self.event_queue = event_queue
         self.max_reorg_depth = max_reorg_depth
         self.last_fetched_block_number = start_block_number - 1
+        self._setup_tenacity()
+
+    def _setup_tenacity(self):
+        # we do not use the decorator syntax, because we do want to
+        # bind this to our instance local logger.
```
jannikluhn: I think without the decorator syntax this could be really confusing. I wouldn't look in `__init__` for anything overwriting a method. Could we either:

- rename the `_rpc_*` functions to something like `_try_rpc_*_once`, so that the `retry` method has at least a different name
- implement a custom `retry` decorator wrapped around the tenacity one, but uses the local logger

?

schmir-at-bb: I think your first suggestion makes it even worse, since the reader will now even think that the method is missing.
I will look into the second option.

schmir-at-bb: I've updated the code to explicitly use tenacity's Retrying object. Please have a look at it again.

schmir-at-bb:
> Not what I had in mind, but also fine with me.
> 
> For the record, here's what I meant:
> 
> ```python
> def retry_local_logger(method):
>     def decorated_method(self, *args, **kwargs):
>         local_retry = retry(..., before_sleep=before_sleep_log(self.logger, logging.WARN)
>         return local_retry(method)(self, *args, **kwargs)
>     return decorated_method
> ```

I would have probably used that if you posted it before. What's nice about doing the retry stuff explicitly is that the caller may decide to retry or not retry method/function calls. Though we don't need that in the event fetcher.

PR #330, opened 2019-08-19
Implement validator balance watcher

### LINKED ISSUE #328: handle the 'not enough eth to send transaction' case
bridge validators may not have enough eth to send a transaction. This is especially true when they are new validators.
At the moment we do have a function, that watches the validator status and only starts confirming transfers when the bridge-client is allowed to validate transactions. At that point the bridge client may not have any eth to send a transaction. One solution would be to also wait until the address has enough eth to send transactions.

jannikluhn:
Closes #328 

- regularly check balance of validator account and emit a warning if it gets too low
- also check the balance at launch and refuse to start if it is too low

I did not go with the solution suggested in #328 as we would only learn that something is wrong when we become a validator. If we'd start the node early we'd likely miss it.

I did not write tests for the watcher as all it does is print log messages which I wouldn't know how to test for in a nice way. I did not write tests for the initial check as we don't have tests for the main function in general, running it against `eth_tester` is not possible, and deploying a new chain for this is overkill.

### DIFF
--- a/tools/bridge/bridge/config.py
+++ b/tools/bridge/bridge/config.py
@@ -12,6 +12,7 @@ from eth_utils import (
     is_checksum_address,
     is_hex,
     to_canonical_address,
+    to_wei,
 )
 from eth_utils.toolz import merge
 
@@ -98,6 +99,8 @@ OPTIONAL_CONFIG_ENTRIES_WITH_DEFAULTS: Dict[str, Any] = {
     "foreign_chain_max_reorg_depth": 10,
     "foreign_chain_event_poll_interval": 5,
     "foreign_chain_event_fetch_start_block_number": 0,
+    "balance_warn_threshold": to_wei(0.1, "ether"),
+    "balance_warn_poll_interval": 60,
 }
 
 CONFIG_ENTRY_VALIDATORS = {
@@ -117,6 +120,8 @@ CONFIG_ENTRY_VALIDATORS = {
     "foreign_bridge_contract_address": validate_checksum_address,
     "foreign_chain_event_fetch_start_block_number": validate_non_negative_integer,
     "validator_private_key": validate_private_key,
+    "balance_warn_threshold": validate_non_negative_integer,
+    "balance_warn_poll_interval": validate_positive_float,
 }
 
 assert all(key in CONFIG_ENTRY_VALIDATORS for key in REQUIRED_CONFIG_ENTRIES)
--- a/tools/bridge/bridge/config.py
+++ b/tools/bridge/bridge/config.py
@@ -99,7 +99,9 @@ OPTIONAL_CONFIG_ENTRIES_WITH_DEFAULTS: Dict[str, Any] = {
     "foreign_chain_max_reorg_depth": 10,
     "foreign_chain_event_poll_interval": 5,
     "foreign_chain_event_fetch_start_block_number": 0,
-    "balance_warn_threshold": to_wei(0.1, "ether"),
+    # disable type check as type hint in eth_utils is wrong, (see
+    # https://github.com/ethereum/eth-utils/issues/168)
+    "balance_warn_threshold": to_wei(0.1, "ether"),  # type: ignore
     "balance_warn_poll_interval": 60,
 }
 
--- a/tools/bridge/bridge/main.py
+++ b/tools/bridge/bridge/main.py
@@ -12,6 +12,7 @@ import click
 import gevent
 import gevent.pool
 from eth_keys.datatypes import PrivateKey
+from eth_utils import from_wei, to_checksum_address
 from gevent.queue import Queue
 from toml.decoder import TomlDecodeError
 from web3 import HTTPProvider, Web3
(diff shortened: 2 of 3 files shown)


schmir-at-bb:
> LGTM, please fix the mypy issue. You can merge this as is.
> 
> #328 isn't completely solved though. At the moment it only logs a warning. I would like to be sure that it stops sending confirmations when the balance is too low and especially that it doesn't try to confirm transactions when the balance is zero..this will be the first thing a new validator sees.

Maybe it is. I didn't see this in the code, but your description says it refuses to start. Does that mean the program aborts?

jannikluhn:
> Maybe it is. I didn't see this in the code, but your description says it refuses to start. Does that mean the program aborts?

It does one check at startup and aborts if the balance is too low. If that check passes though and the balance decreases over time below the threshold, it will just keep running and log warnings.

> I would like to be sure that it stops sending confirmations when the balance is too low and especially that it doesn't try to confirm transactions when the balance is zero.

My initial thought was that in that case the confirmation sender should just retry, so that it automatically succeeds once the account is funded again. Transactions with too little balance will just turn out to be invalid, so I don't see a problem with sending them to the node. I guess the issue is just that we're still gonna send them even if we wouldn't have to anymore because the transfer is already confirmed?

schmir-at-bb:
> Does that mean the program aborts?

Yes, it does abort.

To be clear: This is a bad first user experience. One of the first things the validator is confronted with is a crashing program. We should just wait until he has enough eth (which will be the case when he validated his first block).

Please change the code accordingly. The way to do it is like we did it in the validator status watcher.

jannikluhn:
> To be clear: This is a bad first user experience. One of the first things the validator is confronted with is a crashing program. We should just wait until he has enough eth (which will be the case when he validated his first block).

I think this is dangerous, as it's easy to miss a warning, so some people will just never validate anything. A crashing (or rather safely aborting) program is harder to miss. It's not necessarily the case that they will have enough ETH after they mined their first block because they might have configured a different coinbase address (which is a good idea to do, and IMO should be the default). So this state might persist for a while (until someone tells them something is off).

jannikluhn:
> please fix the mypy issue.

I've disabled the failing check because the type hint in eth_utils it's checking against is wrong. See https://github.com/ethereum/eth-utils/issues/168

schmir-at-bb:
> I think this is dangerous, as it's easy to miss a warning, so some people will just never validate anything. A crashing (or rather safely aborting) program is harder to miss. It's not necessarily the case that they will have enough ETH after they mined their first block because they might have configured a different coinbase address (which is a good idea to do, and IMO should be the default). So this state might persist for a while (until someone tells them something is off).

From what I can tell no one is using a different coinbase address. We do not have any instructions for that. Can you please open a detailed issue in the project repo for that?

To me aborting the program is as I already said just a bad first user experience. The people that don't look into their logs will be the same that don't recognize that their program is aborting.

@saschagoebel, @bernbohmeier, @weilbith 
What's you opinion on this? Should we abort the bridge-validator program when we don't have enough eth to pay for transactions or should we wait until the account has enough eth?

At the moment we also do not abort, when we start the bridge while we're not yet a validator. Instead we wait until the address has become a validator.  We could also wait until the account has enough eth. In the default setup this will happen, when the validator validates his first block.

saschagoebel:
Personally I'd prefer waiting and logging continuously (every few minutes or so) what's wrong, e.g. "Waiting to start because address xyz is not a validator"
or
"Waiting to start because address xyz has not enough funds to send transactions"

The app could then just go back into waiting state (or restart, if that's easier) on too few funds.

Just my $0.02

weilbith:
> The people that don't look into their logs will be the same that don't recognize that their program is aborting.

I agree with that.

> At the moment we also do not abort, when we start the bridge while we're not yet a validator. Instead we wait until the address has become a validator. We could also wait until the account has enough eth. In the default setup this will happen, when the validator validates his first block.

This feels like a good solution to me and makes the program more consistent from my point of view.

I just talked with Jannik and he raised a good argument as I think. It was about the integration into the quickstart script. It is intended to run this script before becoming a validator and there even no balance. It would be super bad to setup and start the bridge validator and then let it immediately die. Would look like misconception for me as user.

As far as I know does the validator right check only happen in the beginning and as soon as the valdiator is activated, the greenlet sets the flag in the planner and terminates. It would be nice, if it would continue and also unset the flag if the validator gets removed after 6 months. The advantage in my eyes is that we can reuse it for the balance check and unset the flag during periods the account has no balance and continue confirming as soon as enough Ether has been received. :man_shrugging:

schmir-at-bb:
> As far as I know does the validator right check only happen in the beginning and as soon as the valdiator is activated, the greenlet sets the flag in the planner and terminates. It would be nice, if it would continue and also unset the flag if the validator gets removed after 6 months. 

We already do that, but we shutdown the program when a validator looses his status as validator.

saschagoebel:
If it then restarts and waits, that would get the job done.

Issue #311, opened 2019-08-13
bridge: Initialize loggers on the module level


schmir-at-bb:
We should initialize the loggers on the module level instead of putting them into some class.

weilbith:
I'm confused. I just pulled the `master` and searched for `logging.getLogger()`. Except the `event_fetcher` all module initialize their logger at the top level. And the event fetcher is intended to be like that.
@schmir-at-bb can this issue be closed? Was it part of a bigger PR and has been forgotten or..?

weilbith:
@schmir-at-bb :bell:

schmir-at-bb:
I removed some of them already, but in the meantime jannik introduced more of them.

PR #334, opened 2019-08-20
Bridge validator - use logger at module level exclusively

weilbith:
This PR is not complete yet. Even if it does "fix" all wrong usages at the moment, I would suggest to keep it open for a short while to also correct occurrences in active PRs etc.

Will close: #311


schmir-at-bb:
I've merged it. We don't have any other PRs open and I like have that in master.

PR #336, opened 2019-08-20
Confirmation sender improvements

schmir-at-bb:
This contains some improvements for the confirmation sender's Watcher class. It retries web3 RPC calls and the queue handling has been simplified.

### DIFF
--- a/tools/bridge/bridge/confirmation_sender.py
+++ b/tools/bridge/bridge/confirmation_sender.py
@@ -93,14 +93,12 @@ class Sender:
         while True:
             transfer_event = self.transfer_event_queue.get()
             assert isinstance(transfer_event, AttributeDict)
-
-            transaction = self.prepare_confirmation_transaction(transfer_event)
+            nonce = self.get_next_nonce()
+            transaction = self.prepare_confirmation_transaction(transfer_event, nonce)
             assert transaction is not None
             self.send_confirmation_transaction(transaction)
 
-    def prepare_confirmation_transaction(self, transfer_event):
-        nonce = self.get_next_nonce()
-
+    def prepare_confirmation_transaction(self, transfer_event, nonce: int):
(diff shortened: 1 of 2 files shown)


tools/bridge/bridge/confirmation_sender.py
```diff
@@ -59,6 +60,12 @@ def run(self):
                 greenlet.kill()
 
 
+sender_retry = tenacity.retry(
```
jannikluhn: A little late, but I've noticed there's also a [middleware](https://web3py.readthedocs.io/en/stable/middleware.html#httprequestretry) provided by web3 that retries requests. There doesn't seem to be any configuration options, but at least it would be used automatically if we enable it once. It's also possible to implement [custom middlewares](https://web3py.readthedocs.io/en/stable/internals.html#internals-middlewares).

schmir-at-bb: At the moment I see no good reason to switch. The nice thing about tenacity is that it would also work with a redis client (if  @saschagoebel succeeds with sneaking in redis) :)

jannikluhn: We could have a tenacity retry middleware:

```Py
def tenacity_retry_middleware(make_request, w3):
    make_request_retry = retry(make_request)
    def middleware(method, params):
        return make_request_retry(method, params)
    return middleware

w3.middleware_onion.add(tenacity_retry_middleware)
```

The main advantage would be that it would be used automatically anytime we use the web3 object. So we wouldn't need all those custom RPC request methods.

schmir-at-bb: I like to separate functions doing RPC requests from those that don't as much as possible, e.g. b924731c6f12e75a15a954aafd27beefd726e6c2
Having it explicit helps when refactoring such stuff.

PR #339, opened 2019-08-21
Last safety net

schmir-at-bb:
This is based on PR #336, please only review the last commit.

This protects against confirming transfers that were not sent to the foreign bridge. We immediately stop the program in that case.

### DIFF
--- a/tools/bridge/bridge/confirmation_sender.py
+++ b/tools/bridge/bridge/confirmation_sender.py
@@ -93,14 +93,12 @@ class Sender:
         while True:
             transfer_event = self.transfer_event_queue.get()
             assert isinstance(transfer_event, AttributeDict)
-
-            transaction = self.prepare_confirmation_transaction(transfer_event)
+            nonce = self.get_next_nonce()
+            transaction = self.prepare_confirmation_transaction(transfer_event, nonce)
             assert transaction is not None
             self.send_confirmation_transaction(transaction)
 
-    def prepare_confirmation_transaction(self, transfer_event):
-        nonce = self.get_next_nonce()
-
+    def prepare_confirmation_transaction(self, transfer_event, nonce: int):
         transfer_hash = compute_transfer_hash(transfer_event)
         transaction_hash = transfer_event.transactionHash
         amount = transfer_event.args.value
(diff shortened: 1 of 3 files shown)


tools/bridge/bridge/confirmation_sender.py
```diff
@@ -19,6 +20,23 @@
...
 
+def make_sanity_check_transfer(foreign_bridge_contract_address):
+    """return a function that checks the final transfer right before it is confirmed
+
+    This is the last safety net. We do check that the transfer has
+    been sent to the foreign_bridge_contract_address here.
+    """
+    assert is_checksum_address(foreign_bridge_contract_address)
+
+    def sanity_check_transfer(transfer_event):
```
jannikluhn: I like this check. Wondering if we should extend it to also compare the transaction we would send against the event (i.e. if the value and the receiver address matches)?

tools/bridge/bridge/confirmation_sender.py
```diff
@@ -19,6 +20,23 @@
...
+        if not isinstance(transfer_event, AttributeDict):
+            raise ValueError("not an AttributeDict")
+        if transfer_event.args["to"] != foreign_bridge_contract_address:
```
jannikluhn: maybe better to use `is_same_address` from `eth_utils`, so that we don't accidentally think two addresses are different, even though they are just differently encoded (hex, checksum, bytes).

schmir-at-bb: I did think about that, but decided against it. That's why there's the assert is_checksum_address().

tools/bridge/bridge/confirmation_sender.py
schmir-at-bb: We can certainly add more checks. I wanted to add this after I broke the event filters in the event fetcher.
Other things that also come to mind: protect against Transfers originating from the wrong contract.

tools/bridge/bridge/confirmation_sender.py
schmir-at-bb: I don't care enough to change it. If you want to have it changed, please open a PR.

PR #340, opened 2019-08-21
Low balance pausing

### LINKED ISSUE #328: handle the 'not enough eth to send transaction' case

jannikluhn:
Pause the confirmation sender if the balance gets too low. The minimum balance can be configured, but has a default of 0.04TLC, which corresponds to a little more than 10 confirmation transactions at the default gas limit.

To limit the number of callbacks, I've introduced a controller queue that the validator status watcher and the validator balance watcher use to push updates to a controller greenlet, which then turns the confirmation task planner on and off accordingly (we've discussed something like this before, but now I think it's worth the additional queue.).

Closes #328

### DIFF
--- a/tools/bridge/bridge/confirmation_task_planner.py
+++ b/tools/bridge/bridge/confirmation_task_planner.py
@@ -14,24 +14,20 @@ class ConfirmationTaskPlanner:
     def __init__(
         self,
         sync_persistence_time: float,
+        minimum_balance: int,
+        control_queue: Queue,
         transfer_event_queue: Queue,
         home_bridge_event_queue: Queue,
         confirmation_task_queue: Queue,
     ) -> None:
-        self.recorder = TransferRecorder()
+        self.recorder = TransferRecorder(minimum_balance)
         self.sync_persistence_time = sync_persistence_time
 
+        self.control_queue = control_queue
         self.transfer_event_queue = transfer_event_queue
         self.home_bridge_event_queue = home_bridge_event_queue
 
         self.confirmation_task_queue = confirmation_task_queue
-
-    def start_validating(self) -> None:
-        self.recorder.start_validating()
-
-    @property
-    def is_validating(self) -> bool:
-        return self.recorder.is_validating
 
     def run(self):
         logger.debug("Starting")
@@ -39,6 +35,7 @@ class ConfirmationTaskPlanner:
             greenlets = [
                 gevent.spawn(self.process_transfer_events),
                 gevent.spawn(self.process_home_bridge_events),
+                gevent.spawn(self.process_control_events),
             ]
             gevent.joinall(greenlets, raise_error=True)
         finally:
@@ -66,6 +63,11 @@ class ConfirmationTaskPlanner:
             else:
                 logger.debug("Received home bridge event")
                 self.recorder.apply_proper_event(event)
+
+    def process_control_events(self) -> None:
+        while True:
+            event = self.control_queue.get()
+            self.recorder.apply_control_event(event)
 
     def check_for_confirmation_tasks(self) -> None:
         confirmation_tasks = self.recorder.pull_transfers_to_confirm()
--- a/tools/bridge/bridge/transfer_recorder.py
+++ b/tools/bridge/bridge/transfer_recorder.py
@@ -1,6 +1,8 @@
-from typing import Dict, List, Set
+import logging
+from typing import Dict, List, Optional, Set
 
 from eth_typing import Hash32
+from eth_utils import from_wei
 from web3.datastructures import AttributeDict
 
 from bridge.constants import (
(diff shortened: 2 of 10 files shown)


jannikluhn:
I also removed the balance check at startup.

jannikluhn:
Updated as discussed, and without tests.

tools/bridge/bridge/confirmation_task_planner.py
```diff
@@ -14,31 +14,28 @@ class ConfirmationTaskPlanner:
...
         self,
         sync_persistence_time: float,
+        minimum_balance: int,
+        control_queue: Queue,
         transfer_event_queue: Queue,
         home_bridge_event_queue: Queue,
         confirmation_task_queue: Queue,
     ) -> None:
-        self.recorder = TransferRecorder()
+        self.recorder = TransferRecorder(minimum_balance)
```
schmir-at-bb: I think we might eventually better pass a TransferRecorder instance instead of adapting the signature every time the transfer recorder needs a new argument.But for the moment we can also keep it like it is.

tools/bridge/bridge/confirmation_task_planner.py
```diff
@@ -67,6 +64,11 @@ def process_home_bridge_events(self) -> None:
                 logger.debug("Received home bridge event")
                 self.recorder.apply_proper_event(event)
 
+    def process_control_events(self) -> None:
+        while True:
+            event = self.control_queue.get()
+            self.recorder.apply_control_event(event)
+
```
schmir-at-bb: As I said, I think all of the process event functions can be unified. But for now that's also ok.

tools/bridge/bridge/control_events.py
```diff
@@ -0,0 +1,17 @@
+from abc import ABC
+
+import attr
+
+
+class ControlEvent(ABC):
+    pass
+
```
schmir-at-bb: I would probably rename that to Event. The FetcherReacherHeadEvent could probably also be moved here, but this one should not end up on the control queue!

tools/bridge/bridge/transfer_recorder.py
```diff
@@ -49,6 +58,25 @@ def apply_proper_event(self, event: AttributeDict) -> None:
...
+        elif isinstance(event, BalanceCheck):
+            balance_sufficient_before = self.is_balance_sufficient
+            self.balance = event.balance
+            balance_sufficient_now = self.is_balance_sufficient
+
+            if balance_sufficient_now and not balance_sufficient_before:
+                logger.info("Account balance is sufficiently high")
+            elif not balance_sufficient_now and balance_sufficient_before:
+                logger.info("Account balance is below safe minimum")
+
```
schmir-at-bb: please raise an error if you don't know what to do with the event

tools/bridge/bridge/transfer_recorder.py
```diff
@@ -49,6 +58,25 @@ def apply_proper_event(self, event: AttributeDict) -> None:
...
+        elif isinstance(event, BalanceCheck):
+            balance_sufficient_before = self.is_balance_sufficient
+            self.balance = event.balance
+            balance_sufficient_now = self.is_balance_sufficient
+
+            if balance_sufficient_now and not balance_sufficient_before:
+                logger.info("Account balance is sufficiently high")
+            elif not balance_sufficient_now and balance_sufficient_before:
+                logger.info("Account balance is below safe minimum")
```
schmir-at-bb: Please log that with logger.warn and with the balance. Always log it when a new event comes in to make it clear something is wrong.

tools/bridge/bridge/validator_balance_watcher.py
```diff
@@ -3,17 +3,25 @@
...
+        w3,
+        validator_address,
+        poll_interval,
+        balance_warn_threshold,
+        control_queue,
     ) -> None:
         self.w3 = w3
         self.validator_address = validator_address
         self.poll_interval = poll_interval
         self.balance_warn_threshold = balance_warn_threshold
```
schmir-at-bb: Please get rid of the `balance_warn_threshold` field. Just get the value and put it on the queue.

schmir-at-bb: Also remove the warning below.

tools/bridge/bridge/transfer_recorder.py
schmir-at-bb: I think that comment applies to where the dispatching is done

schmir-at-bb:
tests are still broken, please fix

Issue #301, opened 2019-08-12
Release docker image for bridge


berndbohmeier:
Build Docker Image on circleci and upload it to dockerhub. We should probably use the same setup as for the tlbc-testnet (i.e. not allow everyone to make a new release and require approval)
We should also be able to release the bridge without releasing a new tlbc-testnet release (i.e. we will probably use a different branch than the branch we use for tlbc-testnet) and vice-versa

weilbith:
The `Dockerfile` does already exist. @schmir-at-bb handles the `tlbc-testnet` image at the moment if I'm correct. What do we need for this issues except pushing a build image?

berndbohmeier:
I dont know what you mean by ralf handles the testnet image. 
We have to write the code for circle ci to build and upload it. You can look at the existing code for the tlbc image and have a chat with Ralf on how to solve that we can release them independently.

weilbith:
**Question:**
On which branches should the new _Docker_ image be published to _DockerHub_?
The `quickstart` as well as the `tlbc-testnet-next` images do so for `master` and `pre-release`.  The "production" image `tlbc-testnet` image only gets tagged and published on the `release` branch.

I have now a version which build and publishes the `tlbc-bridge-validator` image when pushing to the new proposed `release-bridge-validator` branch (super creative I know). I'm not sure how much sense this makes, but do we like to have a version which also deploys on `master`? `pre-release` is related to the `release` branch of the `tlbc-testnet` image. Since the build should be independent, the `tlbc-bridge-validator` should be not listen on this branch. Having a `pre-release-bridge-validator` branch does not make sense to me, since there is nothing to present and agree on with externals. Doesn't it?

schmir-at-bb:
I would create two branches bridge/pre-release and bridge/release and upload from master and those two branches.

master -> bridge-next:master
bridge/pre-release -> bridge-next:pre-release
bridge/release -> bridge:latest

I would not use the validator suffix.

weilbith:
> I would not use the validator suffix.

I called it like this, because other targets are also named so (e.g. `mypy-bridge-validator`) and because I think some additional specificity can't be bad here. But this is not important and I remove it during the additions. :smiley:

weilbith:
Okay I have extended if for Ralf his suggested branch to image name/tag mapping.
I gave up to make it pretty. The only really new part is the cleaning of the branch name since we use now the prefix for the releases. 
[...]

PR #342, opened 2019-08-21
Upload bridge validator docker image with circleci

weilbith:
After all the _CircleCI_ extensions are just re-using existing jobs. This kinda bloats the configuration, but keep it clear and avoids mistakes on optimizations (which aren't even working well).
The release strategy follows the suggestion in #301. The only thing that has changed is the cleaning of the branch name for the docker tag. Since we have now multiple release branches which can be prefixed (e.g. `bridge/pre-release`), this does not work as it is anymore (`trustlines/bridge-next:bridge/pre-release` is not validly tagged).
Moreover I have adjusted the docker compose setup to make use of this image for a production setup. This required a restructuring of the `.yaml` files.

Close: #301


weilbith:
@saschagoebel I assigned you as first reviewer, because Ralf has a lot to do and you could verify at least the docker-compose and README part.

weilbith:
Yes and I think it should be restricted to the release branch only. On any other this does not matter. Pushing a pre-release to satisfy the tests sound awkward to me.

schmir-at-bb:
Can we please test this on the bridge/pre-release and bridge/release branch?(i.e. you should push to those branches).

Which repos do I have to create on dockerhub?

weilbith:
@schmir-at-bb yes of cause.

> Which repos do I have to create on dockerhub?

- `bridge-next`
- `bridge`

This are the names you suggested :wink: I'm open for renaming. 
But do you need to create them first? Doesn't pushing with a new name just initialize a repository? :thinking:
Ping me if they exist, then I push to the other branches. :+1:

schmir-at-bb:
Let's merge this. We still need to figure out what to do with the image diffing...

PR #345, opened 2019-08-22
Make FetcherReachedHeadEvent an Event

schmir-at-bb:
Also add some more fields, like the block time and if it originated
from the home or the foreign chain. Add an enum ChainRole for that.

### DIFF
--- a/tools/bridge/bridge/events.py
+++ b/tools/bridge/bridge/events.py
@@ -1,4 +1,5 @@
 from abc import ABC
+from enum import Enum
 
 import attr
 from web3.datastructures import AttributeDict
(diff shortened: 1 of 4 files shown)


tools/bridge/bridge/events.py
```diff
@@ -13,6 +14,18 @@ class Event(ABC):
 Event.register(AttributeDict)
 
 
+class ChainRole(Enum):
+    home = "home"
```
jannikluhn: Usually enum entries are `UPPER_CAPS` in Python, no?

tools/bridge/bridge/events.py
```diff
@@ -13,6 +14,18 @@ class Event(ABC):
...
+    foreign = "foreign"
+
+
+@attr.s(auto_attribs=True)
+class FetcherReachedHeadEvent:
```
jannikluhn:
```suggestion
class FetcherReachedHeadEvent(Event):
```

schmir-at-bb: right. I missed that.

tools/bridge/bridge/events.py
schmir-at-bb: I don't know. I think it's ugly, but I don't dare to look it up.

PR #348, opened 2019-08-23
Sighup reload logging config

schmir-at-bb:
This is based on PR #346 

Reload the logging configuration on SIGHUP

### DIFF
--- a/tools/bridge/bridge/main.py
+++ b/tools/bridge/bridge/main.py
@@ -224,6 +224,23 @@ def stop(pool, timeout):
         os._exit(os.EX_SOFTWARE)
 
 
+def reload_logging_config(config_path):
+    gevent.getcurrent().name = "sighup"
+    logger.info(
+        f"Handling sighup, trying to reload the logging configuration from {config_path}"
+    )
+    try:
+        config = load_config(config_path)
+        configure_logging(config)
+        logger.info("Logging has been reconfigured")
+    except Exception as err:
+        # this is a signal handler. make sure we don't die as this
+        # will raise the error in the main greenlet.
+        logger.warning(
+            f"Error while trying to reload the logging configuration from {config_path}: {err}"
+        )
+
+
 @click.command()
 @click.option(
     "-c",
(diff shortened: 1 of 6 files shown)


tools/bridge/bridge/main.py
```diff
@@ -224,6 +224,23 @@ def stop(pool, timeout):
...
+        f"Handling sighup, trying to reload the logging configuration from {config_path}"
+    )
+    try:
+        config = load_config(config_path)
+        configure_logging(config)
+        logger.info("Logging has been reconfigured")
+    except Exception as err:
+        # this is a signal handler. make sure we don't die as this
+        # will raise the error in the main greenlet.
+        logger.warning(
```
jannikluhn: Isn't `warning` too light here? `logger.exception` sounds like the more appropriate handler.

tools/bridge/bridge/main.py
```diff
@@ -224,6 +224,23 @@ def stop(pool, timeout):
...
+    logger.info(
+        f"Handling sighup, trying to reload the logging configuration from {config_path}"
+    )
+    try:
+        config = load_config(config_path)
+        configure_logging(config)
+        logger.info("Logging has been reconfigured")
+    except Exception as err:
+        # this is a signal handler. make sure we don't die as this
+        # will raise the error in the main greenlet.
```
jannikluhn: I don't understand this comment (I'm missing some required background though). Maybe clarify with `... don't die so that the error is raised in the main greenlet`, or `don't die as this would raise ...`, depending on what's true.

tools/bridge/bridge/main.py
schmir-at-bb: I don't want to see a traceback in that case. I think a warning is fine. If you change the logging configuration, you will look at the log output anyway. Maybe I'll change it to logger.critical

schmir-at-bb:
> Looks fine, just wondering why we need this? Is it common practice to do this on `SIGHUP`?

Reloading a configuration file on SIGHUP is common practice. I hope this will prove useful, when working with a running tlbc-bridge that somehow misbehaves.

PR #352, opened 2019-08-23
More logging enhancements

schmir-at-bb:
This is based on PR #351, please only review the last 2 commits.

This gives better logging when a signal is received, and logs the internal state periodically or when SIGUSR1 is received.

### DIFF
--- a/tools/bridge/bridge/main.py
+++ b/tools/bridge/bridge/main.py
@@ -224,6 +224,23 @@ def stop(pool, timeout):
         os._exit(os.EX_SOFTWARE)
 
 
+def reload_logging_config(config_path):
+    gevent.getcurrent().name = "sighup"
+    logger.info(
+        f"Handling sighup, trying to reload the logging configuration from {config_path}"
+    )
+    try:
+        config = load_config(config_path)
+        configure_logging(config)
+        logger.info("Logging has been reconfigured")
+    except Exception as err:
+        # this is a signal handler. make sure we don't die as this
+        # will raise the error in the main greenlet.
+        logger.warning(
+            f"Error while trying to reload the logging configuration from {config_path}: {err}"
+        )
+
+
 @click.command()
 @click.option(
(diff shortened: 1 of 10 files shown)


tools/bridge/bridge/main.py
```diff
@@ -241,6 +238,21 @@ def reload_logging_config(config_path):
...
+        gevent.getcurrent().name = name
+        logger.info(f"Received {signal.Signals(signum).name} signal.")
+        f(*args, **kwargs)
+
+    gevent.signal(signum, handler)
+
+
+def log_internal_state(recorder):
+    while True:
+        gevent.sleep(60.0)
```
jannikluhn: Probably nice to have this configurable or at least a constant in `constants.py`..

tools/bridge/bridge/transfer_recorder.py
```diff
@@ -33,6 +33,23 @@ def __init__(self, minimum_balance: int) -> None:
         self.is_validator: Optional[bool] = None
         self.balance: Optional[int] = None
 
+    def log_current_state(self):
+        if self.is_validator:
+            validator_status = "validating"
+        else:
+            validator_status = "not validating"
+
+        balance_in_eth = (self.balance or 0) / 10 ** 18
```
jannikluhn: If I remember correctly, balance is `None` if it hasn't been checked yet. Printing `0` then could be confusing in that case. This probably doesn't happen too often, but it may if the node is offline.

schmir-at-bb: That's right. I didn't care enough to fix this. If the node is offline, the program will not start at the moment. So, this shouldn't be a problem in practice.

tools/bridge/bridge/main.py
schmir-at-bb: I guess I will make this configurable then, maybe, sometime in the future.

Issue #347, opened 2019-08-23
Make prepare_confirmation_transaction RPC call free


schmir-at-bb:
It does call into web3 in order to determine the chainid. I previously assumed it would not do that and consequently it's not wrapped with a retry decorator.

PR #353, opened 2019-08-23
Make prepare_confirmation_transaction RPC free

schmir-at-bb:
This previously did web3 RPC because it needed the chainId. We now
pass that in from the caller.

This comes with a test that uses a web3 middleware, which disallows
RPC requests.

See https://github.com/trustlines-protocol/blockchain/issues/347

### DIFF
--- a/tools/bridge/tests/test_confirmation_sender.py
+++ b/tools/bridge/tests/test_confirmation_sender.py
@@ -130,7 +130,7 @@ def test_transaction_preparation(
     transfer_event,
 ):
     signed_transaction = confirmation_sender.sender.prepare_confirmation_transaction(
-        transfer_event, nonce=1234
+        transfer_event, nonce=1234, chain_id=1122
     )
     transaction = rlp.decode(
         bytes(signed_transaction.rawTransaction), SpuriousDragonTransaction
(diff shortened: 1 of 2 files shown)


tools/bridge/tests/test_confirmation_sender.py
```diff
@@ -151,7 +171,9 @@ def test_transaction_sending(
     validator_address,
 ):
     transaction = confirmation_sender.sender.prepare_confirmation_transaction(
-        transfer_event, nonce=confirmation_sender.sender.get_next_nonce()
+        transfer_event,
+        nonce=confirmation_sender.sender.get_next_nonce(),
+        chain_id=int(w3_home.net.version),
```
berndbohmeier: @schmir-at-bb This will most likely not work with the current parity version we use. Please have a chat with Come about it.

schmir-at-bb: @cducrest Can you please take a look at this commit?

What I'm doing here is signing a transaction. I get the chainid initially with
```
self.chain_id = int(self.w3.net.version)
```
and provide that as a chainid, when signing a transaction.

This looks like it's related to the PR you created for web3, where the validate_chain_id has been changed to compare the chainid against web3.eth.chainId instead of web3.net.version.

I did have web3.eth.chainId in the beginning, but I saw a deprecation warning about that (probably suggesting to use web3.net.version).

At the moment I think I should have kept using web3.eth.chainId as chainid.

cducrest: Afaik, `web3.net.chainId` should give a deprecation warning since it was actually wrong (it was the network id) and it is replaced with `web3.net.version`. You should use `web3.eth.chainId`, and this should not give you a deprecation warning.

I am too lazy to pull your PR to try.

schmir-at-bb: Thanks for the explanation.

schmir-at-bb: This will be fixed with commit 15613e5d79bea965944989e4337351a
(which still has to be merged). I didn't go through adding the fix in this PR, because it would result in merge conflicts.

berndbohmeier: good for me

PR #357, opened 2019-08-26
Replace confirmation sender

schmir-at-bb:
This is based on PR #353.  Please review only the last commit

Replace ConfirmationSender with Sender

The ConfirmationSender has been removed and the Sender class has been
renamed to ConfirmationSender, the Watcher class has been renamed to
ConfirmationWatcher.

They do not share w3 objects now, see
https://github.com/trustlines-protocol/blockchain/issues/304

The confirmation_sender_with_non_validator_account hasn't been used
and is now also removed.

### DIFF
--- a/tools/bridge/bridge/confirmation_sender.py
+++ b/tools/bridge/bridge/confirmation_sender.py
@@ -15,7 +15,7 @@ from bridge.constants import (
     HOME_CHAIN_STEP_DURATION,
 )
 from bridge.contract_validation import is_bridge_validator
-from bridge.service import Service, run_services
+from bridge.service import Service
 from bridge.utils import compute_transfer_hash
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


tools/bridge/bridge/confirmation_sender.py
```diff
@@ -142,6 +105,8 @@ def send_confirmation_transactions(self):
             assert transaction is not None
             self.send_confirmation_transaction(transaction)
 
+    run = send_confirmation_transactions
```
jannikluhn: It seems like we have now two "public" ways to run this: Calling `run` and running all of `self.services`. Can't we just get rid of `run`? As far as I can see, it's only used in tests.

tools/bridge/bridge/confirmation_sender.py
```diff
@@ -230,6 +199,8 @@ def watch_pending_transactions(self):
             receipt = self.wait_for_transaction(oldest_pending_transaction)
             self._log_txreceipt(receipt)
 
+    run = watch_pending_transactions
```
jannikluhn: see above

tools/bridge/bridge/confirmation_sender.py
schmir-at-bb: I would then rather get rid of send_confirmations_transactions and watch_pending_transactions (i.e. replace them with run) 

Is that fine for you?

PR #359, opened 2019-08-26
Bridge end2end tests

berndbohmeier:
This adds a first simple end2end test for the bridge, which will just
test that in a setup with all validators online the transfer will be
completed.
However, this also should function as the blueprint for future end2end
tests. Please pay attention to how I tried to make the tests not flaky
by
  - Test a service to be "up" before proceeding
  - Having timeout tests for fast tests but non flakiness
Please make sure to keep those properties. So most most importantly do not use simple `time.sleep()` as those will either make the tests really slow, or flaky. 

How to run the tests:
you need `parity` and the `tlbc-bridge` installed, then simple run the tests with `pytest`

TODO/known issues. 

- Refactor the code into different files
- Let the tests be run by Circle CI
- Document how to run the tests
- Add uptest for the bridge
- Add more test cases (should be done in a different PR)
    - Less validators
    - What happens if restarted/stopped, backend node stopped and so on
    - ...
- On my machine for some reason parity has to always be killed with `SIGKILL`(automatically down by the tests), please let me know if that is the case for you too? There will be a warning in the test output if that happens. 

So if you want to take over please let me know what you will be working on. Ask me if something is unclear, I will be able to answer.

### DIFF
--- a/tools/bridge/end2end-tests/tests/test_transfer.py
+++ b/tools/bridge/end2end-tests/tests/test_transfer.py
@@ -1,0 +1,427 @@
+import os
+import subprocess
+import time
+import warnings
+from subprocess import Popen
+
+import pytest
+import requests
+from deploy_tools import deploy_compiled_contract
+from deploy_tools.deploy import wait_for_successful_transaction_receipt
+from web3 import HTTPProvider, Web3
+from web3.contract import Contract
+from web3.middleware import construct_sign_and_send_raw_middleware
+
+PARITY_DEV_ACCOUNT = "0x00a329c0648769A73afAc7F9381E08FB43dBEA72"
+PARITY_DEV_KEY = "0x4D5DB4107D237DF6A3D58EE5F70AE63D73D7658D4026F2EEFD2F204C81682CB7"
+
+HOME_RPC_URL = "http://localhost:8545"
+FOREIGN_RPC_URL = "http://localhost:8645"
+
+START_WEI_PER_ACCOUNT = 1 * 10 ** 18
+BRIDGE_FUNDS = 1_000_000 * 10 ** 18
+
+
+class TimeoutException(Exception):
+    pass
+
+
+class Timer:
+    def __init__(self, timeout):
+        self.start_time = None
+        self.timeout = timeout
+
+    def start(self):
+        self.start_time = time.time()
+
+    def is_timed_out(self):
+        if self.start_time is None:
+            raise ValueError("Timer is not started yet")
+        return self.time_passed > self.timeout
+
+    @property
+    def time_passed(self):
+        return time.time() - self.start_time
+
+
+def assert_within_timeout(check_function, timeout, poll_period=0.5):
+    """
+    Runs a check_function with assertions and give it some time to pass
+    :param check_function: The function which will be periodically called. It should contain some assertions
+    :param timeout: After this timeout non passing assertion will be raised
+    :param poll_period: poll interval to check the check_function
+    """
+    timer = Timer(timeout)
+    timer.start()
+
+    while True:
+        try:
+            check_function()
+        except AssertionError as e:
+            if not timer.is_timed_out():
+                time.sleep(poll_period)
+            else:
+                raise TimeoutException(
+                    f"Assertion did not pass after {timeout} seconds. See causing exception for more details."
+                ) from e
+        else:
+            break
+
+
+class Service:
+    def __init__(
+        self,
+        args,
+        *,
+        name=None,
+        env=None,
+        uptest_function=None,
+        timeout=5,
+        poll_interval=0.2,
+        process_settings=None,
+    ):
+        self.args = args
+        self.name = name
+        self.env = env
+        self.timeout = timeout
+        self.poll_interval = poll_interval
+        self._uptest_function = uptest_function
+        self.process = None
+        self._process_settings = process_settings
+
+        if self._process_settings is None:
+            self._process_settings = {}
+
+    def start(self):
+        """Starts the service and wait for it to be up """
+        self.process = Popen(self.args, env=self.env, **self._process_settings)
+        self._wait_for_up()
+        return self.process
+
+    def is_up(self):
+        """Use the uptest to determine if the service is up"""
+        if self._uptest_function is not None:
+            return self._uptest_function
+        else:
+            return True
+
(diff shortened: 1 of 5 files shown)


tools/bridge/end2end-tests/tests/test_transfer.py
```diff
@@ -0,0 +1,427 @@
...
+    """
+    timer = Timer(timeout)
+    timer.start()
+
+    while True:
+        try:
+            check_function()
+        except AssertionError as e:
+            if not timer.is_timed_out():
+                time.sleep(poll_period)
```
jannikluhn: Nitpick:
```suggestion
                time.sleep(max(poll_period, timer.time_left))
```

where `time_left` is a to-be-added property. Otherwise we might not time out even if we should.

tools/bridge/end2end-tests/tests/test_transfer.py
```diff
@@ -0,0 +1,427 @@
...
+        return self.process
+
+    def is_up(self):
+        """Use the uptest to determine if the service is up"""
+        if self._uptest_function is not None:
+            return self._uptest_function
+        else:
+            return True
+
+    def _wait_for_up(self):
```
jannikluhn: can't we replace this with `assert_within_timeout(self.is_up, self.timeout, self.poll_interval)`?

tools/bridge/end2end-tests/tests/test_transfer.py
```diff
@@ -0,0 +1,427 @@
...
+                break
+
+    def terminate(self):
+        try:
+            self.process.terminate()
+            self.process.wait(timeout=10)
+        except subprocess.TimeoutExpired:
+            warnings.warn(f"{self.name} did not terminate in time and had to be killed")
+            self.process.kill()
+            self.process.wait()
```
jannikluhn: Shouldn't we put this in a timeout as well?

tools/bridge/end2end-tests/tests/test_transfer.py
```diff
@@ -0,0 +1,427 @@
...
+        else:
+            return True
+
+    def _wait_for_up(self):
+        start_time = time.time()
+        while True:
+            is_up = self.is_up()
+
+            if not is_up:
+                if time.time() - start_time > self.timeout:
```
jannikluhn: Maybe call `terminate` here as well?

tools/bridge/end2end-tests/tests/test_transfer.py
```diff
@@ -0,0 +1,427 @@
...
+        try:
+            self.process.terminate()
+            self.process.wait(timeout=10)
+        except subprocess.TimeoutExpired:
+            warnings.warn(f"{self.name} did not terminate in time and had to be killed")
+            self.process.kill()
+            self.process.wait()
+
+
+class Node(Service):
```
jannikluhn: Maybe "MainChainNode" or "ForeignChainNode" or at least "ForeignNode" as just "Node" doesnt distinguish the two chains.

tools/bridge/end2end-tests/tests/test_transfer.py
```diff
@@ -0,0 +1,427 @@
...
+
+
+def mine_blocks(web3, number_of_blocks):
+    """Mines `number_of_blocks` blocks. Works by sending out transactions, assumes the parity InstaSeal engine"""
+
+    block_height = web3.eth.blockNumber
+    while web3.eth.blockNumber < block_height + number_of_blocks:
+        web3.eth.sendTransaction({"from": PARITY_DEV_ACCOUNT})
+
+    assert web3.eth.blockNumber == block_height + number_of_blocks
```
jannikluhn: I guess if there are two transactions sent simultaneously, this assert could fail. So maybe just use >=?

tools/bridge/end2end-tests/tests/test_transfer.py
berndbohmeier: Mmh, it is not the same, because in `assert_within_timeout` I try to handle an `AssertionError`. But yeah it is similar, so maybe we could combine them with a little bit of work.

tools/bridge/end2end-tests/tests/test_transfer.py
berndbohmeier: yes, you are right, somewhere should be a terminate in cause the program is not up. But it would probably be better in `start`

tools/bridge/end2end-tests/tests/test_transfer.py
berndbohmeier: Why? A `SIGKILL` will just kill the program

tools/bridge/end2end-tests/tests/test_transfer.py
berndbohmeier: But it is so far just a node? It can be a  foreign or homenode, depending on the config.

berndbohmeier: See line 212

tools/bridge/end2end-tests/tests/test_transfer.py
berndbohmeier: This check is exactly to check this, as this function should mine exactly the number of blocks. This function only works if there are no other transaction at the same time. 
We could change it to `mine_min_blocks` or so and remove the check.

tools/bridge/end2end-tests/tests/test_transfer.py
berndbohmeier: done

tools/bridge/end2end-tests/tests/test_transfer.py
berndbohmeier: I reused the timer

tools/bridge/end2end-tests/tests/test_transfer.py
berndbohmeier: Added a timeout

tools/bridge/end2end-tests/tests/test_transfer.py
berndbohmeier: I call terminate in start if the setup failed.

tools/bridge/end2end-tests/tests/test_transfer.py
berndbohmeier: Changed it to `mine_min_blocks`

Issue #354, opened 2019-08-26
Allow usage of keystore and password file for bridge client


weilbith:
Related to the #302

- [x] Add options to define a keystore and password file.
- [x] Use both to decrypt the private key for the validator.
- [x] Make sure that at least one of both approaches (currently directly define private key as String) is configured.
- [x] No user interaction must be required (don't ask for password)

weilbith:
Feature is complete and all related PRs are merged.

PR #362, opened 2019-08-27
Feature/bridge client keystore option

weilbith:
Related to: #354


schmir-at-bb:
My nitpick wasn't that clear. Feel free to merge, but consider using full paths next time

weilbith:
> We may think about better names for the keys in the config file. Those that we have in here were the first that came to my mind, when we talked about the issue. But that shouldn't keep us from merging this PR.

I'm pretty fine with the naming. I think you intentionally put all information into the names. When I read them I don't miss anything. So I can't come up with better suggestions and would keep them.

> My nitpick wasn't that clear. Feel free to merge, but consider using full paths next time

No I really like this hint. It makes sense to me and if we already talk about it, I think we should apply it to have a clean code base when merging.

---

If CI runs through I go for the merge. I don't expect any trouble. Afterwards I will rebase PR #363.

PR #363, opened 2019-08-27
Feature/bridge client nested environment variables

weilbith:
This is partially based on #362 
Please review/merge that PR first and focus on the last two commits only.
I do not feel 100% confident with this solution. More in general with the continuous support of environment variables for hierarchical options. Anyways I think using section <> dicts seem the best approach to handle more complex options like the private key. Therefore I hope this solution is good enough. Rather than only support the specific configuration sections, this should be a stable overall solution. Thereby also the logging can be configured via environment variables as well.


weilbith:
@schmir-at-bb we are on the same page. I've implemented it just to continue the support for the current functionality. Without the environment configuration would need always a mix with a TOML file. But at least after writing this PR, I vote for dropping the support and go for the TOML file as single source.
So as you stated I merge it to have a full functional master again and close #354. But I would love to see this stuff to become removed again at a later point.

Issue #361, opened 2019-08-27
bridge: fix broken nonce handling 


schmir-at-bb:
The nonce handling is broken. I get these messages in the log, when multiple transfers are being confirmed:
```
2019-08-27 15:08:14,722 WARNING  [send-confirmation-transactions] bridge.confirmation_sender: Retrying bridge.confirmation_sender.ConfirmationSender._rpc_send_raw_transaction in 5.0 seconds as it raised {'code': -32010, 'message': 'Transaction nonce is too low. Try incrementing the nonce.'}.
```

We should fix that and also not retry on those errors and rather get a new nonce.

berndbohmeier:
One hint here, parity has a special function for the nonce which includes all pending transaction. We use this also in the relay server .

berndbohmeier:
Might not be the issue here, but could be.

schmir-at-bb:
Thanks. For the record, this is what we use in the relay server:

```
    def get_tx_infos(self, user_address, block_identifier="pending"):
        if self.is_parity and block_identifier == "pending":
            nonce = int(
                self._web3.manager.request_blocking("parity_nextNonce", [user_address]),
                16,
            )
        else:
            nonce = self._web3.eth.getTransactionCount(
                user_address, block_identifier=block_identifier
            )
        return TxInfos(
            balance=self._web3.eth.getBalance(
                user_address, block_identifier=block_identifier
            ),
            nonce=nonce,
            gas_price=self.fetch_gas_price(),
        )
```

schmir-at-bb:
We now also call parity_nextNonce and retry with a new nonce in case we get a 'nonce too low' error.

PR #364, opened 2019-08-27
Fix nonce handling

schmir-at-bb:
This is base on PR #360, please only review the last 3 commits.

This fixes the nonce handling: #361

### DIFF
--- a/tools/bridge/bridge/confirmation_sender.py
+++ b/tools/bridge/bridge/confirmation_sender.py
@@ -15,7 +15,7 @@ from bridge.constants import (
     HOME_CHAIN_STEP_DURATION,
 )
 from bridge.contract_validation import is_bridge_validator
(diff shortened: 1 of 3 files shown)


tools/bridge/bridge/confirmation_sender.py
```diff
@@ -80,30 +89,65 @@ def __init__(
             )
         ]
 
-    @sender_retry
+        self.is_parity = self.w3.clientVersion.startswith("Parity")
```
jannikluhn: should we just raise a `SetupException` if it's not Parity? It's the only client that works for the trustlines chain.

schmir-at-bb: This would break at least the tests. Maybe others want to use the bridge in a different setup. I wouldn't want to prevent them doing that.

PR #366, opened 2019-08-28
Enhance transfer recorder

schmir-at-bb:
This enhance the transfer recorder a bit. It confirms transfers in the order they happened on the foreign chain and it ignores Transfer events with zero balance or zero recipient.

Also I've added some basic tests for the transfer recorder. I've only noticed afterwards the we do have some in `test_confirmation_task_planner`. We should probably move them to `test_transfer_recorder.py`

### DIFF
--- a/tools/bridge/bridge/transfer_recorder.py
+++ b/tools/bridge/bridge/transfer_recorder.py
@@ -11,7 +11,7 @@ from bridge.constants import (
     TRANSFER_EVENT_NAME,
 )
 from bridge.events import BalanceCheck, Event, FetcherReachedHeadEvent, IsValidatorCheck
-from bridge.utils import compute_transfer_hash
+from bridge.utils import compute_transfer_hash, sort_events
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


tools/bridge/bridge/transfer_recorder.py
```diff
@@ -87,12 +91,18 @@ def pull_transfers_to_confirm(self) -> List[AttributeDict]:
...
+        sort_events(confirmation_tasks)
         return confirmation_tasks
 
     def apply_proper_event(self, event: AttributeDict) -> None:
         event_name = event.event
 
         if event_name == TRANSFER_EVENT_NAME:
+            if event.args.value == 0 or is_same_address(
+                event.args["from"], ZERO_ADDRESS
+            ):
```
jannikluhn: Do we actually want the balance check? It would be a nice way to test if the bridge is working without having tokens. And for attackers it doesn't really make a difference if they send 0 or 1 wei.

the "from" address will never be zero, so maybe an `assert` is more appropriate.

schmir-at-bb: HomeBridge.sol will reject these transfers (I accidentally created one with zero amount). So, yes, we want the balance check.
Our TrustlinesNetworkToken.sol has an emit Transfer with the zero address as sender in it's mint function. So, I would also like to filter those out.

Issue #358, opened 2019-08-26
End2end tests findings


berndbohmeier:
I will just write down some of my findings, to discuss if we can improve them and then create tasks to do so. 

- We will not do this. I did not catch the problem that I have to wait for the reorg depth. One of our goals was to make the bridge more talk to find problems more easy, so maybe we should log every event already on "latest" or even "pending" with the note that we will not react to it now, but wait for the reorg depth. 

- Deferred ~The confirmations of the bridge were failing because the bridge had no funds, but there was no logging of the bridge that the transaction were failing. I dont know the exact reason why this is not logged, as I can see some code that should do this I think? We should make sure that this is logged.~ This was also because of the reorg depth. Same applies as above maybe we want to log it already before? I dont know. 

- [x] We could also check that the bridge has enough funds? But we probably do not need it as we will fund the bridge only once and this is more a problem that arises in testing. 

- [x] Some of the services use the block time as poll interval, we should make those poll intervals also configurable. We can of course use the block time still as the default. 

- [x] To make the tests more reliable it would be good to have a way to test if the bridge is "up" whatever that means, maybe you have an idea how to do that.

schmir-at-bb:
> I will just write down some of my findings, to discuss if we can improve them and then create tasks to do so.
> 
>     * I did not catch the problem that I have to wait for the reorg depth. One of our goals was to make the bridge more talk to find problems more easy, so maybe we should log every event already on "latest" or even "pending" with the note that we will not react to it now, but wait for the reorg depth.

We don't even fetch them at the moment. I'm not sure if it makes sense, I would rather not do this at the moment. 

If this becomes a problem in a real world setup, it's because the chain doesn't produce new blocks or we don't receive any. This is something I would rather try to determine and log.

> 
>     * ~The confirmations of the bridge were failing because the bridge had no funds, but there was no logging of the bridge that the transaction were failing. I dont know the exact reason why this is not logged, as I can see some code that should do this I think? We should make sure that this is logged.~ This was also because of the reorg depth. Same applies as above maybe we want to log it already before? I dont know.

I would log the fact that we're waiting for more blocks if we see a transaction that's not yet reorg safe. We have that information already (at least for the first transaction that we're waiting for)

> 
>     * We could also check that the bridge has enough funds? But we probably do not need it as we will fund the bridge only once and this is more a problem that arises in testing.

Then let's at least abort the program if the bridge has zero funds when starting tlbc-bridge.

> 
[...]

PR #369, opened 2019-08-28
Add a webservice providing information about the internal state

schmir-at-bb:
This may prove useful for the end2end tests, see the discussion in
https://github.com/trustlines-protocol/blockchain/issues/358

It can be enabled by the following snippet:

  [webservice]
  enabled = true
  host = "127.0.0.1"
  port = 9500

The '/bridge/internal-state' endpoint provides information about the
internal state. It looks like the following:

,----
| % http http://localhost:9500/bridge/internal-state
| HTTP/1.1 200 OK
| Date: Wed, 28 Aug 2019 22:12:37 GMT
| content-length: 1058
| content-type: application/json
|
| {
|     "bridge": {
|         "config": {
|             "foreign_bridge_contract_address": "0xb4c79daB8f259C7Aee6E5b2Aa729821864227e84",
|             "foreign_chain_event_fetch_start_block_number": 0,
|             "foreign_chain_max_reorg_depth": 0,
|             "foreign_chain_token_contract_address": "0x731a10897d267e19B34503aD902d0A29173Ba4B1",
|             "foreign_rpc_url": "http://localhost:9200",
|             "home_bridge_contract_address": "0x771434486a221c6146F27B72fd160Bdf0eb1288e",
|             "home_chain_event_fetch_start_block_number": 3849548,
|             "home_chain_max_reorg_depth": 2,
|             "home_rpc_url": "http://localhost:9100"
|         },
|         "process": {
|             "gid": 1001,
|             "loadavg": [
|                 0.44,
|                 0.41,
|                 0.32
|             ],
|             "pid": 14478,
|             "uid": 1001
|         },
|         "recorder": {
|             "balance": "13286498470000000000",
|             "is-balance-sufficient": true,
|             "is-validating": true,
|             "is-validator": true,
|             "last-fetcher-reached-head-event": [
|                 {
|                     "chain-role": "foreign",
|                     "last-fetched-block-number": 5673,
|                     "timestamp": 1567030356.144592
|                 },
|                 {
|                     "chain-role": "home",
|                     "last-fetched-block-number": 3868357,
|                     "timestamp": 1567030356.4055734
|                 }
|             ],
|             "num-completions": 0,
|             "num-scheduled-hashes": 0,
|             "num-transfer-events": 0
|         },
|         "version": "0.1.0"
|     }
| }
`----

I've chosen the falcon framework (https://falconframework.org/)
instead of Flask. This looks much simpler and doesn't require any
magic, like thread local variables. Also, it has no dependencies. I
actually also considered bottle and pyramid.

The keys inside the response are kebab-cased (except those in the
config section, because they correspond to the config entries in the
TOML file). This was a deliberate decision. Please don't try to change
my mind.

### DIFF
--- a/tools/bridge/bridge/config.py
+++ b/tools/bridge/bridge/config.py
@@ -121,6 +121,13 @@ def validate_private_key(private_key: Any) -> dict:
     return private_key
 
 
+def validate_webservice(webservice_dict: Dict) -> Dict:
+    """validate the webservice dictionary"""
+    if not isinstance(webservice_dict, dict):
+        raise ValueError("webservice must be a dictionary")
+    return webservice_dict
+
+
 def validate_logging(logging_dict: Dict) -> Dict:
     """validate the logging dictionary
 
@@ -156,9 +163,11 @@ OPTIONAL_CONFIG_ENTRIES_WITH_DEFAULTS: Dict[str, Any] = {
     # disable type check as type hint in eth_utils is wrong, (see
     # https://github.com/ethereum/eth-utils/issues/168)
     "minimum_validator_balance": to_wei(0.04, "ether"),  # type: ignore
+    "webservice": {},
 }
 
 CONFIG_ENTRY_VALIDATORS = {
+    "webservice": validate_webservice,
     "logging": validate_logging,
     "home_rpc_url": validate_rpc_url,
     "home_rpc_timeout": validate_non_negative_integer,
--- a/tools/bridge/bridge/main.py
+++ b/tools/bridge/bridge/main.py
@@ -38,6 +38,7 @@ from bridge.service import Service, start_services
 from bridge.utils import get_validator_private_key
 from bridge.validator_balance_watcher import ValidatorBalanceWatcher
 from bridge.validator_status_watcher import ValidatorStatusWatcher
+from bridge.webservice import DummyWebservice, InternalState, Webservice
 
 logger = logging.getLogger(__name__)
 
@@ -234,6 +235,40 @@ def make_validator_balance_watcher(config, control_queue):
     )
 
 
+public_config_keys = (
+    "foreign_rpc_url",
+    "home_rpc_url",
+    "foreign_chain_max_reorg_depth",
+    "home_chain_max_reorg_depth",
+    "foreign_chain_token_contract_address",
+    "foreign_bridge_contract_address",
+    "home_bridge_contract_address",
+    "foreign_chain_event_fetch_start_block_number",
+    "home_chain_event_fetch_start_block_number",
+)
+
+
+def make_webservice(*, config, recorder):
+    d = config["webservice"]
+    if d["enabled"]:
+        ws = Webservice(host=d["host"], port=d["port"])
+    else:
+        ws = DummyWebservice()
+
+    def encode_address(v):
+        if isinstance(v, bytes):
+            return to_checksum_address(v)
+        else:
+            return v
+
+    public_config = {k: encode_address(config[k]) for k in public_config_keys}
+
+    ws.enable_internal_state(
+        InternalState(recorder=recorder, public_config=public_config)
+    )
+    return ws
+
+
 def stop(pool, timeout):
     logger.info("Stopping...")
 
@@ -353,6 +388,7 @@ def main(ctx, config_path: str) -> None:
     )
 
     validator_balance_watcher = make_validator_balance_watcher(config, control_queue)
+    webservice = make_webservice(config=config, recorder=recorder)
 
     services = (
         [
@@ -373,6 +409,7 @@ def main(ctx, config_path: str) -> None:
         + sender.services
         + watcher.services
         + confirmation_task_planner.services
+        + webservice.services
     )
 
     install_signal_handler(
--- a/tools/bridge/bridge/main.py
+++ b/tools/bridge/bridge/main.py
@@ -263,9 +263,7 @@ def make_webservice(*, config, recorder):
 
     public_config = {k: encode_address(config[k]) for k in public_config_keys}
 
-    ws.enable_internal_state(
-        InternalState(recorder=recorder, public_config=public_config)
-    )
+    ws.enable_internal_state(InternalState(recorder=recorder, config=public_config))
     return ws
 
 
--- a/tools/bridge/bridge/main.py
+++ b/tools/bridge/bridge/main.py
@@ -38,7 +38,7 @@ from bridge.service import Service, start_services
 from bridge.utils import get_validator_private_key
 from bridge.validator_balance_watcher import ValidatorBalanceWatcher
 from bridge.validator_status_watcher import ValidatorStatusWatcher
-from bridge.webservice import DummyWebservice, InternalState, Webservice
+from bridge.webservice import InternalState, Webservice
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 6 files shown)


tools/bridge/bridge/config.py
```diff
@@ -121,6 +121,13 @@ def validate_private_key(private_key: Any) -> dict:
     return private_key
 
 
+def validate_webservice(webservice_dict: Dict) -> Dict:
+    """validate the webservice dictionary"""
+    if not isinstance(webservice_dict, dict):
+        raise ValueError("webservice must be a dictionary")
+    return webservice_dict
```
jannikluhn: We should also check here that if it's enabled a host and port is given and that there are no additional entries.

tools/bridge/bridge/transfer_recorder.py
```diff
@@ -33,6 +39,28 @@ def __init__(self, minimum_balance: int) -> None:
...
         self.balance: Optional[int] = None
+        self.last_fetcher_reached_head_event: Dict[
+            ChainRole, FetcherReachedHeadEvent
+        ] = {}
+
+    def get_state_summary(self):
+        return {
+            "is-validator": self.is_validator,
+            "balance": str(self.balance or -1),
+            "num-transfer-events": len(self.transfer_events),
```
jannikluhn: maybe also include the transfer events directly, could be helpful to track a specific transfer

tools/bridge/bridge/webservice.py
```diff
@@ -0,0 +1,89 @@
...
+<h1>Welcome to tlbc-bridge</h1>
+</header>
+
+<p>
+You have reached tlbc-bridge's REST server. This is only meant for
+debugging.
+</p>
+
+<p>
+If you see this in a production setup, please remove the
```
jannikluhn: What's wrong with using this in production? Could be useful for monitoring.

tools/bridge/bridge/transfer_recorder.py
```diff
@@ -33,6 +39,28 @@ def __init__(self, minimum_balance: int) -> None:
 
         self.is_validator: Optional[bool] = None
         self.balance: Optional[int] = None
+        self.last_fetcher_reached_head_event: Dict[
+            ChainRole, FetcherReachedHeadEvent
+        ] = {}
+
+    def get_state_summary(self):
```
jannikluhn: Conceptually, I think this would fit better in the `InternalState`. Then the job of the `InternalState` would be to present what the application is doing and the job of the `TransferRecorder` is actually do stuff. Right now the `TransferRecorder` has to do both. Also, in the future we could add other data sources that the `InternalState` can read from and then we'd have presentation logic spread over the whole application.

tools/bridge/bridge/main.py
```diff
@@ -234,6 +235,40 @@ def make_validator_balance_watcher(config, control_queue):
...
+    "home_chain_event_fetch_start_block_number",
+)
+
+
+def make_webservice(*, config, recorder):
+    d = config["webservice"]
+    if d["enabled"]:
+        ws = Webservice(host=d["host"], port=d["port"])
+    else:
+        ws = DummyWebservice()
```
jannikluhn: this isn't really a webservice at all. Calling `enable_internal_state` on it does nothing and is confusing. Returning `None` if no webserver should run would be much cleaner.

tools/bridge/bridge/main.py
```diff
@@ -234,6 +235,40 @@ def make_validator_balance_watcher(config, control_queue):
     )
 
 
+public_config_keys = (
```
jannikluhn: Why can't all config keys be public?

tools/bridge/bridge/main.py
```diff
@@ -234,6 +235,40 @@ def make_validator_balance_watcher(config, control_queue):
...
+    else:
+        ws = DummyWebservice()
+
+    def encode_address(v):
+        if isinstance(v, bytes):
+            return to_checksum_address(v)
+        else:
+            return v
+
+    public_config = {k: encode_address(config[k]) for k in public_config_keys}
```
jannikluhn: Not really happy with the `encode_address` function. What do you think about changing the way we deal with the config file to keep them in human readable form (i.e. not decode addresses in the `validate` functions, but rather in `main.py` where we use them)? Or add a `decode_config` function that transforms a human readable into a machine readable one?

tools/bridge/bridge/transfer_recorder.py
schmir-at-bb: All of them? I probably wouldn't do it in this endpoint, because that could be many transfers and it makes looking at the output a bit harder.
Though we could have another one providing a complete list.

tools/bridge/bridge/config.py
schmir-at-bb: yes, this is a shortcut I took. Contrary to what I said in the daily I'm not sure what is the best way to solve this. Maybe we should just manually check that here.

tools/bridge/bridge/webservice.py
schmir-at-bb: At the moment it wouldn't hurt. I'm not sure in which direction this is going. E.g. if we add the possibility to modify the internal state (or inject new 'faked' transfer events), it's probably better to turn if off.
It's purpose at this stage is mainly to help implementing the end2end tests, so that's why I added that message.

tools/bridge/bridge/main.py
schmir-at-bb: Some of them contain confidential information, like the validators private key.

tools/bridge/bridge/main.py
schmir-at-bb: Yes, this was a quick hack and I'm also not happy with it.

I kind of like the decode_config idea, though I haven't thought it through. 
At this point I wonder if we should use marshmallow for the validation of the configuration.

tools/bridge/bridge/main.py
schmir-at-bb: ok, I will change that.

tools/bridge/bridge/transfer_recorder.py
```diff
@@ -33,6 +39,28 @@ def __init__(self, minimum_balance: int) -> None:
...
+        return {
+            "is-validator": self.is_validator,
+            "balance": str(self.balance or -1),
+            "num-transfer-events": len(self.transfer_events),
+            "num-scheduled-hashes": len(self.scheduled_hashes),
+            "num-completions": len(self.completion_hashes),
```
berndbohmeier: Please really reconsider the use of kebab case. The problem is that in a lot of languages including Python and JavaScript, the minus is not a valid character, so you break the workflow to just convert it to an object and access it via `obj.var` because `obj.in-valid` is invalid. There is a reason the convention of json is camelCase. If not please use at least snake_case. 
I know you said it's a deliberate decision, but a bad deliberate decision is still a bad decision.

tools/bridge/bridge/transfer_recorder.py
schmir-at-bb: I kind of see your point. But, on the other hand we would have to reach into the transfer_recorder from the internal state object.

> Also, in the future we could add other data sources that the InternalState can read from

We will do this.

My suggestion then would be to add a 

```
@functools.singledispatch
def get_internal_state_summary(obj):
```
and implement that for each object which we want to publish via the `InternalState` class. 
That would mean we still spread `get_internal_state_summary` over the whole application, but in a dedicated function and we can implement that for each class that we want to without changing the class. And the `InternalState` only has to call that function and not reach into the other classes.
 
 
Would that be fine for you?

tools/bridge/bridge/transfer_recorder.py
schmir-at-bb: I don't think there is a general convention to use camelCase for json keys (see https://stackoverflow.com/questions/5543490/json-naming-convention). 

If you want to convert it to an object where you can use dot access it would be easy to that by replacing '-' with '_', so I don't buy into your argument.

I can switch to using snake_case however.

PR #372, opened 2019-08-29
Use Marshmallow for configuration

jannikluhn:
What I have not done (yet):

- group config entries: We want to have home chain and foreign chain config groups, so that we can share schemas there. They are not exactly equal though and I didn't find a way to extend schemas.
- validate the logging config: We haven't been doing it so far, but it might be useful. Future PR?


jannikluhn:
now rebased and including webservice config

schmir-at-bb:
> validate the logging config: We haven't been doing it so far, but it might be useful. Future PR?

My guess is that this isn't easy. I would not try it.

schmir-at-bb:
I've pushed some changes to your branch. Please try to use pre-commit. If it doesn't work on your machine, please complain!

schmir-at-bb:
> They are not exactly equal though and I didn't find a way to extend schemas.

Can you use a subclass with additional fields?

jannikluhn:
Yep, that probably works. Wonder how I didn't think of this, sounds quite natural...

PR #374, opened 2019-08-30
Config Grouping

jannikluhn:
Group config values by chain and udpate the readme accordingly.

### DIFF
--- a/tools/bridge/README.md
+++ b/tools/bridge/README.md
@@ -83,47 +83,42 @@ docker build --file ./Dockerfile --tag tlbc-bridge ../../
 The bridge validator client can be configured with a TOML configuration file whose path must be
 given via the `--config` (`-c`) CLI paramter.
 
-The following table lists all available options. Configuration entries
-with a default value are optional.
-
-|                      Name                      |    Default    |                              Description                             |
-|:----------------------------------------------:|:-------------:|:--------------------------------------------------------------------:|
-|                `foreign_rpc_url`               |               |   URL to JSON-RPC endpoint of foreign chain node [HTTP(S) protocol]  |
-|                 `home_rpc_url`                 |               |    URL to JSON-RPC endpoint of home chain node [HTTP(S) protocol]    |
-|              `foreign_rpc_timeout`             |     `180`     |   timeout option of the `web3` _HTTPProvider_ for the foreign chain  |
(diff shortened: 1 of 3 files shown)


tools/bridge/README.md
```diff
@@ -83,47 +83,42 @@ docker build --file ./Dockerfile --tag tlbc-bridge ../../
...
+bridge_contract_address = "0x77E0d930cF5B5Ef75b6911B0c18f1DCC1971589C"  # address of the home bridge contract
+gas_price = 10000000000000000000  # gas price in Wei for confirmation transactions
+minimum_validator_balance = 40000000000000000
+balance_warn_poll_interval = 60.0
+
+[validator_private_key]  # the private key of the validator
+raw = "0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+# or alternatively (but not at the same time)
+keystore_path = "~/validator_keystore.json"
+keystore_password_path = "~/validator_password"
```
schmir-at-bb: Did you test that this actually works? (tilde expansion)

tools/bridge/bridge/config.py
```diff
@@ -110,47 +110,44 @@ def validate_host_and_port_if_enabled(self, in_data, **kwargs):
...
+
+class ForeignChainSchema(ChainSchema):
+    max_reorg_depth = fields.Integer(missing=1, validate=validate_non_negative)
+    token_contract_address = AddressField(required=True)
+
+
+class HomeChainSchema(ChainSchema):
+    max_reorg_depth = fields.Integer(missing=10, validate=validate_non_negative)
+    gas_price = fields.Integer(
+        missing=to_wei(10, "ether"), validate=validate_non_negative
```
schmir-at-bb: 10 ether gas price?

jannikluhn: :scream: fixed

tools/bridge/README.md
jannikluhn: huh, no, I just expected it to work. Turns out it doesn't, maybe something we should fix at some point

tools/bridge/bridge/config.py
schmir-at-bb: Should we check that the gas price is within reasonable bounds? In order to prevent the user from configuring something like we had here?

PR #376, opened 2019-09-01
Add end2end tests to CI

berndbohmeier:


.circleci/config.yml
```diff
@@ -454,6 +454,25 @@ jobs:
...
+      - run:
+          name: Make parity executable
+          command: chmod 755 ~/bin/parity
+      - run:
+          name: Run pytest
+          command: |
+            cd tools/bridge
+            make test-end2end
+            cd ../..
+
```
schmir-at-bb: This could be written as `make -C tools/bridge test-end2end`

ghost (approved):

jannikluhn (approved):

.circleci/config.yml
berndbohmeier: Ah okay, I copied it from the other circle ci code. I changed it everywhere.

PR #379, opened 2019-09-02
Limit pending tx

schmir-at-bb:
- some documentation fixes
- limit number of pending transactions


schmir-at-bb:
> The change looks good to me, but I'm not sure why the number of pending transactions is bound to the reorg depth.
> Wouldn't it be fine to have just a fixed number?

All of the transactions that are in the last max_reorg_depth blocks are considered pending in our implementation. So, if we increase the max_reorg_depth, we need to increase the max_pending_transactions (or we end up with blocks where we can't put any transactions into).

Issue #368, opened 2019-08-28
bridge: wait for chain to have reached the start block before doing sanity checks


schmir-at-bb:
tlbc-bridge should wait for the chain to have reached the start block before it is doing it's sanity checks. Otherwise the program will just abort, which is a bad experience for users setting up the system with the quickstart script.

jannikluhn:
I would say we don't need this for the home chain as everything will be deployed at genesis (at least it could, not sure what our plan is here).

For the foreign chain it could make sense. We could also add another config option `token_contract_deployment_block_number` as `start_block_number` can be even later. But I don't know if it's worth it.

schmir-at-bb:
It was already a problem for me during testing and it won't hurt if we do this for both chains.

schmir-at-bb:
signal handling should still work and we could already start the webservice when waiting for this to happen. Also it should be clear that it's waiting from looking at the logs.

schmir-at-bb:
@jannikluhn  Can you take care of this?

PR #380, opened 2019-09-03
Wait for start block

jannikluhn:
Closes https://github.com/trustlines-protocol/blockchain/issues/368

- Moved the recorder out of the confirmation task planner because we need it for the webserver which we want to start earlier than the planner.
- Wait until the two chains have passed the event fetch start block number and the node has synced these blocks (not sure if the latter is strictly necessary, but shouldn't hurt)
- Only initialize and run everything after that

Unfortunately, this has a problem with cancellations right now: https://github.com/trustlines-protocol/blockchain/compare/trustlines-protocol:2bba5d7...trustlines-protocol:df84f07#diff-f925bfdaf316e7fb4a51e87452ddf410R536

The intention is to only start the main services if the start block greenlet has exited on its own and wasn't killed with a signal. If it's killed I expected `greenlet.successful()` to be false, but this doesn't seem to be the case. Not sure how to solve this, other than by switching to trio.

### DIFF
--- a/tools/bridge/bridge/confirmation_task_planner.py
+++ b/tools/bridge/bridge/confirmation_task_planner.py
@@ -14,14 +14,14 @@ logger = logging.getLogger(__name__)
 class ConfirmationTaskPlanner:
     def __init__(
         self,
+        recorder: TransferRecorder,
         sync_persistence_time: float,
-        minimum_balance: int,
         control_queue: Queue,
         transfer_event_queue: Queue,
         home_bridge_event_queue: Queue,
         confirmation_task_queue: Queue,
     ) -> None:
-        self.recorder = TransferRecorder(minimum_balance)
+        self.recorder = recorder
         self.sync_persistence_time = sync_persistence_time
 
         self.control_queue = control_queue
--- a/tools/bridge/bridge/main.py
+++ b/tools/bridge/bridge/main.py
@@ -36,6 +36,7 @@ from bridge.contract_validation import (
 from bridge.event_fetcher import EventFetcher
 from bridge.events import ChainRole
 from bridge.service import Service, start_services
+from bridge.transfer_recorder import TransferRecorder
 from bridge.utils import get_validator_private_key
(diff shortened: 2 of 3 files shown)


jannikluhn:
(For some reason, my precommits didn't catch the linting error in the previous commit, neither black nor flake8. Not sure if that's on me or if some configuration here is slightly broken).

schmir-at-bb:
> Not sure how to solve this, other than by switching to trio.

read the fine manuals :). I am well aware of the fact that you would have preferred trio.

I've pushed a commit that fixes the check. It's a bit surprising, but the documentation is pretty clear on this. We also relay on the fact that killing a greenlet is considered succesful in main.py's last call to gevent.joinall with raise_error=True.

schmir-at-bb:
> (For some reason, my precommits didn't catch the linting error in the previous commit, neither black nor flake8. Not sure if that's on me or if some configuration here is slightly broken).

Indeed. We have two different configurations for flake8.

The top level setup.cfg does not ignore E501 (line too long), while the one in tools/bridge does ignore it.
flake8 seems to search for setup.cfg based on a common base directory:
```
141 ~/trustlines/protocol/blockchain (git)-[tmp] % git reset --hard df84f07d829d27ec5d0d2061a13cd91c6b5
HEAD is now at df84f07 Wait until synced far enough before doing anything
~/trustlines/protocol/blockchain (git)-[tmp] % flake8 tools/bridge/bridge/main.py                  
~/trustlines/protocol/blockchain (git)-[tmp] % flake8 tools/bridge/bridge/main.py tools/quickstart/setup.py
tools/bridge/bridge/main.py:105:120: E501 line too long (121 > 119 characters)
tools/bridge/bridge/main.py:115:120: E501 line too long (143 > 119 characters)
tools/bridge/bridge/main.py:122:120: E501 line too long (138 > 119 characters)
```

schmir-at-bb:
> Indeed. We have two different configurations for flake8.

PR for that one: https://github.com/trustlines-protocol/blockchain/pull/381

PR #381, opened 2019-09-04
Consolidate flake8 config in top-level .flake8

schmir-at-bb:
We had differing flake8 configs spread in multiple setup.cfg
files. This resulted in pre-commit not detecting errors locally, but
on flagging them on circle ci.


schmir-at-bb:
dammit, flake8 has a strange way to find it's config file.

When running flake8 in the top level directory it picks up, .flake8 from the top level directory.
When running inside e.g. tools/bridge it picks up setup.cfg, which doesn't have a flake8 section. 

Consequently it prints lots of errors.

schmir-at-bb:
This is working now, please review.

PR #394, opened 2019-09-09
Allow using quickstart script to setup non-validator node

### LINKED ISSUE #391: Option to setup an observer node in quickstart script
The quickstart script should ask the user:
"Do you want to set up a validator node(1) or a regular node(2)?" 

If chosen validator, it should ask for key generation and bridge setup and netstats.
If chosen observer it will not generate a key and not start the bridge, but it will still ask for netstats setup.

jannikluhn:
Closes #391 

Haven't fully tested it yet as I need to install docker first.

What confused me a little is that at the moment `/config/user-config.yaml` appears to be used instead of `/config/validator-config.toml`. That's actually what we'd want for this PR, but not for the normal case. I'll check if this is true or if I'm misreading the code when I've setup docker to test this PR.

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -19,7 +19,7 @@ services:
       - ${HOST_BASE_DIR:-.}/trustlines/config:/config/custom
       - ${HOST_BASE_DIR:-.}/trustlines/enode:/config/network
     command: >-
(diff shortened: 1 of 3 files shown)


jannikluhn:
Tested and apparently working.

Issue #385, opened 2019-09-09
quickstart in python wrap up


schmir-at-bb:
There are multiple minor things to do/check.
- [x] pin the quickstart image being used by the quickstart script
- [x] make sure we stop old containers installed with the old quickstart script
- [x] install a static docker executable?

schmir-at-bb:
stop containers: 32e39a1e52a61a066713e0726668f6a4bde21618

PR #396, opened 2019-09-10
Pin quickstart docker image on digest

jannikluhn:
See bullet point 1 in #385 

I removed [this section](https://github.com/trustlines-protocol/blockchain/compare/trustlines-protocol:51c0d04...trustlines-protocol:3b65822#diff-8846dd18c9ee9c09dadeee541156c2b8L64-L74) which I honestly didn't fully understand, but suspect is not needed anymore.

### DIFF
--- a/quickstart.sh
+++ b/quickstart.sh
@@ -76,7 +76,6 @@ EOF
 
   docker run --rm --tty --interactive \
     --volume /var/run/docker.sock:/var/run/docker.sock \
-    --volume /usr/bin/docker:/usr/bin/docker \
     --volume "${PWD}":/data \
     --volume "$DATA_DIR":/quickstart/trustlines \
     $DOCKER_IMAGE_NAME \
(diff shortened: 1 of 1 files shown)


jannikluhn:
Making testing harder is the last thing I want to do. So I just removed the digest and put everything in a single variable so that one only has to set a single variable.

schmir-at-bb:
> I removed [this section](https://github.com/trustlines-protocol/blockchain/compare/trustlines-protocol:51c0d04...trustlines-protocol:3b65822#diff-8846dd18c9ee9c09dadeee541156c2b8L64-L74) which I honestly didn't fully understand, but suspect is not needed anymore.

We used this previously to use the latest images from dockerhub. The code has been reused to use the latest quickstart image. It's not needed anymore.

schmir-at-bb:
Beware: I've force pushed to this branch, since I needed to adapt the tag already (and three commits looked like too much for me)

PR #397, opened 2019-09-10
Stop containers started by the old quickstart script

jannikluhn:
See bullet point 2 in #385

I didn't manage to do this without adding docker itself to the quickstart script container. Using `docker-compose` to stop the containers doesn't work (it doesn't find them -- maybe because they are not defined in the docker-compose file) and `docker` itself appears to be available, but when run crashes due to a missing `libltdl.so` file


schmir-at-bb:
I've already merged that one. The quickstart script still needs to be fixed.

PR #405, opened 2019-09-13
Reconfirm completes transfer

schmir-at-bb:
This is based on PR #404, please only review the last two commits.

Add reconfirmCompletesTransfer view

This view can be used by the validators to check if calling
confirmTransfer a second time would complete the transfer. This makes
sense after validator set changes.

### DIFF
--- a/contracts/contracts/bridge/HomeBridge.sol
+++ b/contracts/contracts/bridge/HomeBridge.sol
@@ -98,6 +98,41 @@ contract HomeBridge {
         }
     }
 
+    // check if a 2nd confirmTransfer would complete a transfer. this
+    // can happen after validator set changes.
+    function reconfirmCompletesTransfer(
+        bytes32 transferHash,
+        bytes32 transactionHash,
+        uint256 amount,
+        address payable recipient
+    ) public view returns (bool) {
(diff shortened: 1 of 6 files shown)


contracts/contracts/bridge/HomeBridge.sol
```diff
@@ -98,6 +98,41 @@ contract HomeBridge {
         }
     }
 
+    // check if a 2nd confirmTransfer would complete a transfer. this
+    // can happen after validator set changes.
+    function reconfirmCompletesTransfer(
```
jannikluhn: I guess it's more a question about Solidity patterns, but wouldn't it be better to have the confirm function return true or false, depending on if it was successful or not instead of having a separate function with separate logic to do the same check?

schmir-at-bb: I think a view function callable by **anyone** is useful. You have to call confirmTransfer as validator. Also it's non obvious what the return value should be.

jannikluhn: I believe you can fake the sender address for calls. But good point about the return value.

PR #409, opened 2019-09-13
Feature/e2e test foreign node start late

cducrest:
closes: Probably some tasks related to having some e2e tests

still WIP, tests tearup does not terminate the bridges properly so tests will not work when ran a second time without `killall tlbc-bridge`

edit: should be good now, I successfully ran tests multiple times in a row and bridges were stopping properly.

### DIFF
--- a/tools/bridge/end2end-tests/tests/test_transfer.py
+++ b/tools/bridge/end2end-tests/tests/test_transfer.py
@@ -18,6 +18,9 @@ PARITY_DEV_KEY = "0x4D5DB4107D237DF6A3D58EE5F70AE63D73D7658D4026F2EEFD2F204C8168
 HOME_RPC_URL = "http://localhost:8545"
 FOREIGN_RPC_URL = "http://localhost:8645"
 
+# How many blocks on the foreign chain must be mined for a transfer to be considered final by the bridge
+REORG_DEPTH = 3
+
 START_WEI_PER_ACCOUNT = 1 * 10 ** 18
 BRIDGE_FUNDS = 1_000_000 * 10 ** 18
 
@@ -89,6 +92,11 @@ def assert_within_timeout(check_function, timeout, poll_period=0.5):
                 break
 
 
+def assert_after_timout(check_function, timeout):
+    time.sleep(timeout)
+    check_function()
+
+
 class Service:
     def __init__(
         self,
@@ -122,9 +130,6 @@ class Service:
         except TimeoutException:
             self.terminate()
             raise
-
-    def is_started(self):
-        return self.process is None
 
     def is_up(self):
         """Determine if the service is up"""
@@ -209,7 +214,7 @@ class Bridge(Service):
             "foreign_chain": {
                 "rpc_url": FOREIGN_RPC_URL,
                 "rpc_timeout": 5,
-                "max_reorg_depth": 3,
+                "max_reorg_depth": REORG_DEPTH,
                 "event_poll_interval": 0.5,
                 "event_fetch_start_block_number": 0,
                 "bridge_contract_address": str(foreign_bridge_contract_address),
@@ -270,7 +275,7 @@ def accounts(accounts):
     return [PARITY_DEV_ACCOUNT] + list(accounts)
 
 
-@pytest.fixture()
+@pytest.fixture(scope="session")
 def account_keys(account_keys):
     return [PARITY_DEV_KEY] + list(account_keys)
 
@@ -392,6 +397,11 @@ def validator_keys(account_keys):
 
 
 @pytest.fixture()
+def bridge_addresses(validators):
+    return validators
+
+
+@pytest.fixture()
 def system_address(accounts):
     """Address pretending to be the system address."""
     return accounts[0]
@@ -504,8 +514,7 @@ def test_simple_transfer(
         ).transact({"from": sender}),
     )
 
-    # reorg depth is 3
-    mine_min_blocks(web3_foreign, 3)
+    mine_min_blocks(web3_foreign, REORG_DEPTH)
 
     def check_balance():
         foreign_balance_after = token_contract.functions.balanceOf(sender).call()
@@ -537,8 +546,7 @@ def test_offline_validators_validates_not_complete_transfer(
         ).transact({"from": sender}),
     )
 
-    # reorg depth is 3
-    mine_min_blocks(web3_foreign, 3)
+    mine_min_blocks(web3_foreign, REORG_DEPTH)
 
     def check_balances_bridge_transfer_not_complete():
         foreign_balance_transfer_not_complete = token_contract.functions.balanceOf(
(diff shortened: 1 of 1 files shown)


tools/bridge/end2end-tests/tests/test_transfer.py
```diff
@@ -85,6 +85,14 @@ def assert_within_timeout(check_function, timeout, poll_period=0.5):
                 break
 
 
+def assert_false_during_timeout(check_function, timeout, poll_period=0.5):
```
berndbohmeier: What is the use of that?

tools/bridge/end2end-tests/tests/test_transfer.py
```diff
@@ -109,13 +117,16 @@ def __init__(
...
     def start(self):
         """Starts the service and wait for it to be up """
-        self.process = Popen(self.args, env=self.env, **self._process_settings)
-        try:
-            self._wait_for_up()
-            return self.process
-        except TimeoutException:
-            self.terminate()
-            raise
+        if self.is_up():
```
berndbohmeier: This check only checks if the service is up. This is not the same as started, so please remove it here.

tools/bridge/end2end-tests/tests/test_transfer.py
```diff
@@ -450,18 +461,37 @@ def bridges(
...
         )
-        bridge.start()
         bridges.append(bridge)
 
-    yield
+    yield bridges
+
+    for bridge in bridges:
+        if bridge.is_up():
+            bridge.terminate()
```
schmir-at-bb: Please just unconditionally terminate them.

tools/bridge/end2end-tests/tests/test_transfer.py
cducrest: To assert something remains false during a period. Similarly to how you want to see an event occur within a time limit, you might want to check that the opposite event did not happen within the same time limit.

tools/bridge/end2end-tests/tests/test_transfer.py
cducrest: Is it not? Since `start()` will wait for the service to be up before returning. If we do no have this and a tests attempt to start a bridge twice, the bridges fixture will not properly terminate all bridges.

tools/bridge/end2end-tests/tests/test_transfer.py
cducrest: What if they are already terminated? Then I have an error at teardown.

tools/bridge/end2end-tests/tests/test_transfer.py
berndbohmeier: So you mean
```
sleep(timeout)
assert
```
Why do you write it so complicated backwards?

tools/bridge/end2end-tests/tests/test_transfer.py
```diff
@@ -391,6 +394,11 @@ def validator_keys(account_keys):
     return account_keys[7:10]
 
 
+@pytest.fixture()
+def bridges_address(validators):
```
berndbohmeier: bridge_addresses?

tools/bridge/end2end-tests/tests/test_transfer.py
```diff
@@ -117,6 +126,9 @@ def start(self):
             self.terminate()
             raise
 
+    def is_started(self):
+        return self.process is None
```
berndbohmeier: Now you added it but do not even use it yourself?
Unuesed code can easily lead to broken code, which is also what happened, it should be
`self.process is not None`

If you dont use it, maybe better remove it.

tools/bridge/end2end-tests/tests/test_transfer.py
```diff
@@ -253,12 +268,12 @@ def mine_min_blocks(web3, number_of_blocks):
         )
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture()
```
berndbohmeier: the do not change, so they can stay session fixtures

tools/bridge/end2end-tests/tests/test_transfer.py
```diff
@@ -485,3 +522,186 @@ def check_balance():
...
+    def check_balances_bridge_transfer_not_complete():
+        foreign_balance_transfer_not_complete = token_contract.functions.balanceOf(
+            sender
+        ).call()
+        home_balance_transfer_not_complete = web3_home.eth.getBalance(sender)
+
+        assert foreign_balance_before - foreign_balance_transfer_not_complete == value
+        assert home_balance_before == home_balance_transfer_not_complete
+
+    assert_within_timeout(check_balances_bridge_transfer_not_complete, 10)
```
berndbohmeier: This should be assert after timeout, otherwise it could be that you checked just to early.

tools/bridge/end2end-tests/tests/test_transfer.py
```diff
@@ -485,3 +522,186 @@ def check_balance():
...
+    def check_balance_transfer_complete():
+        home_balance_transfer_complete = web3_home.eth.getBalance(sender)
+        assert home_balance_transfer_complete - home_balance_before == value
+
+    # We need to wait for the transaction to be completed before starting bridges[2]
+    assert_within_timeout(check_balance_transfer_complete, 10)
+
+    bridges[2].start()
+
+    time.sleep(10)
```
berndbohmeier: same as above. I would write a function `assert_after_timeout` so that you do not always have to write the sleep there.

tools/bridge/end2end-tests/tests/test_transfer.py
```diff
@@ -485,3 +522,186 @@ def check_balance():
...
+
+    inactive_address = accounts[2]
+    assert inactive_address not in bridges_address
+    initial_validators = [bridges_address[0], inactive_address, bridges_address[2]]
+    validator_proxy_contract.functions.updateValidators(initial_validators).transact(
+        {"from": system_address}
+    )
+
+    bridges[0].start()
+    bridges[1].start()
```
berndbohmeier: Maybe a comment that this is not a validator yet.

berndbohmeier:
And can you maybe rebase on master, so that we can see the coverage metrics?

tools/bridge/end2end-tests/tests/test_transfer.py
berndbohmeier: As discussed, this is changed for a is_started check

tools/bridge/end2end-tests/tests/test_transfer.py
cducrest: But then I have to write a whole function to give to the `assert_after_timeout`

berndbohmeier: Yes. You can also leave it like this. I just think it would make it clearer what the sleep is doing there and communicate that we do not want sleeps for no reason

PR #410, opened 2019-09-13
Test config

schmir-at-bb:
Add some tests for bridge.main, bridge.webservice, bridge.config

### DIFF
--- a/tools/bridge/tests/conftest.py
+++ b/tools/bridge/tests/conftest.py
@@ -250,3 +250,14 @@ bridge_contract_address = "0x771434486a221c6146F27B72fd160Bdf0eb1288e"
 [validator_private_key]
 raw = "0xb8dcbb8a564483279579e04bffacbd76f79df157cfbebed84079673b32d9e72f"
 """
+
+
+@pytest.fixture
+def webservice_config():
+    """webservice part of the configuration"""
+    return """
+[webservice]
+enabled = true
+host = "127.0.0.1"
+port = 9500
+"""
--- a/tools/bridge/tests/conftest.py
+++ b/tools/bridge/tests/conftest.py
@@ -1,11 +1,14 @@
 import gevent.monkey
 import gevent.pool
 import pytest
+import toml
 from deploy_tools import deploy_compiled_contract
 from eth_tester import EthereumTester
 from gevent.queue import Queue
 from web3 import EthereumTesterProvider, Web3
 from web3.contract import Contract
+
+import bridge.config
 
 # check if gevent did it's monkeypatching
 if "time" not in gevent.monkey.saved:
@@ -261,3 +264,13 @@ enabled = true
 host = "127.0.0.1"
 port = 9500
 """
+
+
+@pytest.fixture
+def load_config_from_string():
+    """returns a function that loads a configuration dictionary from a string"""
+
+    def load(s):
+        return bridge.config.ConfigSchema().load(toml.loads(s))
+
+    return load
--- a/tools/bridge/tests/conftest.py
+++ b/tools/bridge/tests/conftest.py
@@ -231,3 +231,22 @@ def write_config(tmp_path):
         return str(p)
 
     return write
+
+
+@pytest.fixture
+def minimal_config():
+    """return a minimal configuration"""
+
+    return """
+[foreign_chain]
+rpc_url = "http://localhost:9200"
+token_contract_address = "0x731a10897d267e19B34503aD902d0A29173Ba4B1"
+bridge_contract_address = "0xb4c79daB8f259C7Aee6E5b2Aa729821864227e84"
+
+[home_chain]
+rpc_url = "http://localhost:9100"
+bridge_contract_address = "0x771434486a221c6146F27B72fd160Bdf0eb1288e"
+
+[validator_private_key]
+raw = "0xb8dcbb8a564483279579e04bffacbd76f79df157cfbebed84079673b32d9e72f"
+"""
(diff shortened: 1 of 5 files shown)


tools/bridge/tests/conftest.py
```diff
@@ -218,3 +221,56 @@ def home_bridge_contract(
...
+token_contract_address = "0x731a10897d267e19B34503aD902d0A29173Ba4B1"
+bridge_contract_address = "0xb4c79daB8f259C7Aee6E5b2Aa729821864227e84"
+
+[home_chain]
+rpc_url = "http://localhost:9100"
+bridge_contract_address = "0x771434486a221c6146F27B72fd160Bdf0eb1288e"
+
+[validator_private_key]
+raw = "0xb8dcbb8a564483279579e04bffacbd76f79df157cfbebed84079673b32d9e72f"
+"""
```
jannikluhn: maybe nicer to define it as a dict and then use `toml.dumps`

jannikluhn: if not, you could also use `textwrap.dedent`

tools/bridge/tests/conftest.py
```diff
@@ -218,3 +221,56 @@ def home_bridge_contract(
...
+    return """
+[webservice]
+enabled = true
+host = "127.0.0.1"
+port = 9500
+"""
+
+
+@pytest.fixture
+def load_config_from_string():
```
jannikluhn: Something I've noticed a while ago, this is a just a good example: What's the reason to make this a fixture and not just a regular module-level function? I have the feeling that in general we're using fixtures quite often to just for the convenience of passing them via parameters (most of the time test data) and less often to run proper setup code which I think is the original intention of fixtures. Not sure if this is actually a problem, though, but curious about others' opinions.

tools/bridge/tests/test_config.py
```diff
@@ -0,0 +1,42 @@
...
+level = "INFO"
+
+[logging.loggers.urllib3]
+level = "INFO"
+"""
+
+
+def test_load_minimal_config(write_config, minimal_config):
+    cfg = bridge.config.load_config(write_config(minimal_config))
+    print(cfg)
```
jannikluhn: can be removed, no?

tools/bridge/tests/test_config.py
```diff
@@ -0,0 +1,42 @@
...
+
+    assert cfg["foreign_chain"]["event_fetch_start_block_number"] == 0
+    assert cfg["foreign_chain"]["max_reorg_depth"] >= 10
+
+
+def test_logging_key(write_config, minimal_config):
+    cfg = bridge.config.load_config(
+        write_config(minimal_config + "\n" + example_logging_config)
+    )
+    print(cfg)
```
jannikluhn: ditto

tools/bridge/tests/test_config.py
```diff
@@ -0,0 +1,42 @@
...
+# web3 is too verbose with level debug
+[logging.loggers.web3]
+level = "INFO"
+
+[logging.loggers.urllib3]
+level = "INFO"
+"""
+
+
+def test_load_minimal_config(write_config, minimal_config):
```
jannikluhn: This test looks useless to me. It just checks that the values defined in the test itself are sensible. Am I missing something?

jannikluhn: If we want to have a test for config file loading, I'd suggest to start with some config, dump it, load it, and check that the result is equal. And maybe the same with some invalid values.

tools/bridge/tests/test_config.py
```diff
@@ -0,0 +1,42 @@
...
+    # make sure we have some sensible defaults
+    assert cfg["home_chain"]["gas_price"] >= 10 ** 9
+    assert cfg["home_chain"]["event_fetch_start_block_number"] == 0
+    assert cfg["home_chain"]["max_reorg_depth"] >= 5
+
+    assert cfg["foreign_chain"]["event_fetch_start_block_number"] == 0
+    assert cfg["foreign_chain"]["max_reorg_depth"] >= 10
+
+
+def test_logging_key(write_config, minimal_config):
```
jannikluhn: ditto. In addition, this tests that "version" and "incremental is added which I guess is a useful test, but seems to be more of a side effect. At least its not apparent from the name of the test.

tools/bridge/tests/conftest.py
schmir-at-bb: yes, you're right. I had been thinking about that when I wrote it. One problem is that I didn't have a canonical module to put this into.
Having said that my first implementation of that fixture also used the write_config fixture. So, in that case it had to be a fixture.

The same is also true for the minimal_config fixture, which is just a string. This one previously was just a global variable inside one test module, but became a fixture when I needed it in another test module.
I will try to move some of those into a dedicated test module and see how it looks like (not in this PR).

I would say the other use of fixtures is to parametrize tests.

tools/bridge/tests/test_config.py
schmir-at-bb: I actually like these, especially when a test fails. (but somehow I always have to defend these print calls :)

tools/bridge/tests/test_config.py
schmir-at-bb: > This test looks useless to me. It just checks that the values defined in the test itself are sensible. Am I missing something?

This is one of the tests that found an actual bug! I did change the `foreign_bridge.max_reorg_depth` default after writing it. 

Also, if we add a new config key without defining a default, this test will fail. I think that's useful.

If we dump a config, it will contain the full set of keys.

PR #412, opened 2019-09-16
Add monitor to quickstart script

jannikluhn:
Closes https://github.com/trustlines-network/project/issues/652

The monitor actually doesn't run at the moment because it requires https://github.com/trustlines-protocol/tlbc-monitor/pull/76. So I could only test it up to the point at which it attempts to start the monitor container.


jannikluhn:
I took the liberty to change the path to the shared config files once again: I moved it to the "trustlines" directory so that all files the quickstart script creates are confined to a single directory.

PR #414, opened 2019-09-17
Tests/test for confirmation task planner

### LINKED ISSUE #407: Add unit test for the confirmation task planer
The test coverage claims that there is no unit test coverage for the confirmation task planer.

berndbohmeier:
Closes #407


berndbohmeier:
@schmir-at-bb Any idea why pre-commit betrayed me there again?

schmir-at-bb:
> @schmir-at-bb Any idea why pre-commit betrayed me there again?

we don't run mypy via pre-commit in this repository. we can add it, and should copy the configuration from the contracts repository.

Issue #415, opened 2019-09-18
Parity wrapper should copy chain spec file to certain path


jannikluhn:
Before launching parity, the wrapper should copy the chain spec file to a (configurable? hard coded?) location -- usually a docker volume.

PR #419, opened 2019-09-18
Update parity-wrapper to copy chain spec file to shared volume

cducrest:
Update parity-wrapper to copy chain spec file to shared volume to get spec file out of container
closes: https://github.com/trustlines-protocol/blockchain/issues/415

### DIFF
--- a/docker/parity_wrapper.sh
+++ b/docker/parity_wrapper.sh
@@ -202,7 +202,13 @@ function runParity() {
   exec $PARITY_BIN $PARITY_ARGS
 }
 
+function copySpecFileToSharedVolume() {
+  echo "Copying trustline spec file to shared volume"
+  cp /config/trustlines-spec.json /shared-config/trustlines-spec.json
+}
+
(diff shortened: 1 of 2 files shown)


schmir-at-bb:
or even better yet, something like /shared/

(in case we ever want to share some data directories)

docker/parity_wrapper.sh
```diff
@@ -202,7 +202,13 @@ function runParity() {
   exec $PARITY_BIN $PARITY_ARGS
 }
 
+function copySpecFileToSharedVolume() {
+  echo "Copying trustline spec file to shared volume"
+  cp /config/trustlines-spec.json /config/chain-spec/trustlines-spec.json
```
schmir-at-bb: Why does it still copy to /config/chain-spec/ ?

cducrest: Apparently because I failed commiting because pre-commit complains that I do not have the executable `shfmt`

PR #432, opened 2019-09-23
Release

jannikluhn:
I tested the pre-release with the following steps and did not find any issues:

- Plain `docker run trustlines/tlbc-testnet-next:pre-release`: It runs fine. It shows an error message in the log output, but arguably that's correct behavior and doesn't matter:

```
Copying trustline spec file to shared volume
cp: cannot create regular file '/shared/trustlines-spec.json': No such file or directory
```

- Rebuild the current version of the quickstart image (the one that's pinned in the current version of `quickstart.sh`, commit df645f3d2cc386d90c85bb8f1e38f4f6e6925342), but with the `trustlines/tlbc-testnet-next:pre-release` in the docker compose file. If I run this, again, the same error message, but other than that no issues.


berndbohmeier:
Why did we not build it in a way that it works without failing with an error message? Just checking first if the folder exists should be enough.

schmir-at-bb:
> I dont think we should give an error message when actually every thing is fine, how should someone know that it is okay to have the error message? We should fix that first.

I would just ship the image with an empty /shared directory.

berndbohmeier:
We now check first if shared exist, and skip the cp if the folder does not exist. 
I think this is better, as this way the user can see if he set up something wrong, for example when he accidentally mounted the volume to the wrong folder, he will see a message. 
Are you fine with that @schmir-at-bb

schmir-at-bb:
yes, already approved.

PR #433, opened 2019-09-23
Only copy chain spec if target directory exists

jannikluhn:


docker/parity_wrapper.sh
```diff
@@ -203,8 +204,12 @@ function runParity() {
 }
 
 function copySpecFileToSharedVolume() {
-  echo "Copying trustline spec file to shared volume"
-  cp /config/trustlines-spec.json /shared/trustlines-spec.json
+  if [[ -d "$SHARED_VOLUME_PATH" ]]; then
+    echo "Copying trustline spec file to shared volume"
+    cp /config/trustlines-spec.json "$SHARED_VOLUME_PATH/trustlines-spec.json"
```
berndbohmeier: there is a slash already in `/shared/` is that not an issue?

berndbohmeier: okay, looks like this should not be an issue apparently.

jannikluhn: According to this [Stackoverflow answer](https://stackoverflow.com/a/24026057), multiple slashes are treated as a single one, so this should work. I also tested it and it seems to work.

PR #439, opened 2019-09-25
"Pin" mainnet node versions to stable

### LINKED ISSUE #430: Pin parity version to stable for the mainnet node
We want to be able to follow main chain forks. We need to pin the parity version we are using for the mainnet node to `stable` instead of specific version.
See: https://github.com/trustlines-protocol/blockchain/blob/master/docker/docker-compose.yaml

jannikluhn:
Closes #430 

I've tested the new version by running the bridge end to end test with it. I'm uncertain if we should do this in CI: On the one hand, we regularly test with the newest stable release, on the other hand it makes the tests less deterministic.

### DIFF
--- a/tools/bridge/docker/docker-compose-base.yaml
+++ b/tools/bridge/docker/docker-compose-base.yaml
@@ -1,7 +1,7 @@
 version: "2.4"
 services:
   node_foreign:
-    image: parity/parity:v2.5.5-stable
+    image: parity/parity:stable
     command:
       - "--light"
       - "--no-download"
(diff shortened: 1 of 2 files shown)


tools/bridge/docker/docker-compose-base.yaml
```diff
@@ -1,7 +1,7 @@
 version: "2.4"
 services:
   node_foreign:
-    image: parity/parity:v2.5.5-stable
+    image: parity/parity:stable
```
berndbohmeier: Do you know what this file is used for?

jannikluhn: I'm not entirely sure, but I don't think we use it for anything. Validators could use it though if they don't want to use the quickstart script. Don't think it's strictly necessary though.

PR #454, opened 2019-10-09
Feature/delete confirming validators after complete

cducrest:
Gas costs for a transfer with 50 validators, require 50% to complete transfer, and no validator set changes:

### with optimization:

first confirm: 97_156
other confirm: 82_156
last confirm: 169_808
useless confirm: 28_234

total confirm: 2_156_552

### without optimization:

first confirm: 97_156
other confirm: 82_156
last confirm: 217_578
useless confirm: 28_234

total confirm: 2_204_322

### DIFF
--- a/contracts/contracts/bridge/HomeBridge.sol
+++ b/contracts/contracts/bridge/HomeBridge.sol
@@ -75,7 +75,7 @@ contract HomeBridge {
         if (_confirmTransfer(transferStateId, msg.sender)) {
             // We have to emit the events here, because _confirmTransfer
             // doesn't even receive the necessary information to do it on
-            // it's own
+            // its own
 
             emit Confirmation(
                 transferHash,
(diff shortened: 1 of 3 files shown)


berndbohmeier:
Any idea why the gas cost of a confirm do not change even though we removed:
` transferState[transferStateId].numConfirmations += 1;`?

berndbohmeier:
The mapping and the list belong together, so did you try to clear also the mapping, maybe that also saves gas?

cducrest:
You can't delete a mapping since it does not know which key has an actual set value, so we'd need to loop over all the validators and delete the value of `isConfirmedByValidator[validator]`. This increases gas costs by 100k.

berndbohmeier:
Can you somewhere post the test code?
I mean cleaning a slot should refund 15k, so I dont get why this does not work.

cducrest:
there it is: https://github.com/trustlines-protocol/blockchain/compare/feature/clean-is-confirmed-mapping?expand=1

I don't fully understand how mappings work internally, but afaict in the struct it is just one word, and each bool is on bit. So not so much storage freed by deleting it.

schmir-at-bb:
I'm looking into this.

Issue #452, opened 2019-10-08
Invastigate bridge issue right after fully synced node


berndbohmeier:
After start via the quickstart, the bridge waits for the node to be synced. After is has synced the following error occurs. 
This happens several times, but after a while it seems to be fine. 
```
2019-10-08 14:15:40,949 INFO     [wait_for_start_block] bridge.main: foreign node is synced until block number 1280559 of 1435033, but hasn't reached event fetch start block number 1321331 yet. Waiting...
2019-10-08 14:15:45,958 INFO     [wait_for_start_block] bridge.main: foreign node is synced until block number 1293564 of 1435034, but hasn't reached event fetch start block number 1321331 yet. Waiting...
2019-10-08 14:15:50,965 INFO     [wait_for_start_block] bridge.main: foreign node is synced until block number 1303885 of 1435034, but hasn't reached event fetch start block number 1321331 yet. Waiting...
2019-10-08 14:15:55,973 INFO     [wait_for_start_block] bridge.main: foreign node is synced until block number 1318703 of 1435034, but hasn't reached event fetch start block number 1321331 yet. Waiting...
2019-10-08 14:16:00,982 INFO     [wait_for_start_block] bridge.main: foreign node is synced until block number 1332136 of 1435035 and has passed the event fetch start block number 1321331
2019-10-08 14:16:11,044 ERROR    [main] bridge.main: Application error
Traceback (most recent call last):
  File "/opt/bridge/lib/python3.6/site-packages/bridge/main.py", line 560, in main
    main_services = make_main_services(config, recorder, stop_pool)
  File "/opt/bridge/lib/python3.6/site-packages/bridge/main.py", line 374, in make_main_services
    transfer_event_fetcher = make_transfer_event_fetcher(config, transfer_event_queue)
  File "/opt/bridge/lib/python3.6/site-packages/bridge/main.py", line 215, in make_transfer_event_fetcher
    validate_contract_existence(token_contract)
  File "/opt/bridge/lib/python3.6/site-packages/bridge/contract_validation.py", line 18, in validate_contract_existence
    contract_code = contract.web3.eth.getCode(contract.address)
  File "/opt/bridge/lib/python3.6/site-packages/web3/eth.py", line 137, in getCode
    [account, block_identifier],
  File "/opt/bridge/lib/python3.6/site-packages/web3/manager.py", line 97, in request_blocking
    raise ValueError(response["error"])
[...]

PR #459, opened 2019-10-11
Update bridge so that it waits for the foreign chain to be fully synced

cducrest:
Update bridge so that it waits for the foreign chain to be fully synced before starting transfers.

It seems that when a light node is not fully synced, requests are not properly handled.

I could not find a good way to reduce redundancy in between `wait_for_chain_synced_until_block` and `wait_for_chain_fully_synced_past_block` without having one function that does two different things with a flag, since most of it is logging infos.

closes: https://github.com/trustlines-protocol/blockchain/issues/452

### DIFF
--- a/tools/bridge/bridge/main.py
+++ b/tools/bridge/bridge/main.py
@@ -67,7 +67,7 @@ def configure_logging(config):
     )
 
 
-def wait_for_chain_synced_until(w3, start_block_number, timeout, chain_role):
+def wait_for_chain_synced_until_block(w3, start_block_number, timeout, chain_role):
     retry = tenacity.retry(
         wait=tenacity.wait_exponential(multiplier=1, min=5, max=timeout),
         before_sleep=tenacity.before_sleep_log(logger, logging.WARN),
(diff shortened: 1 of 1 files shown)


schmir-at-bb:
> It seems that when a light node is not fully synced, requests are not properly handled.

If we can detect this error, I think we should just retry the request. We just want to check that the contract exists at that address, not that the chain is synced. I consider the current behaviour a bug in parity, if they fix it, we don't have to wait for the chain to be synced.

schmir-at-bb:
It looks like commit 1b009a7 tries to fix an unrelated issue (#473). As already noted there, I expect it to be fixed in another way. Please remove it from this PR.

Can we distinguish the error we see in #452 from the error we would get from an address with no code deployed?  If yes, we should retry the request instead of waiting for the chain to be fully synced.

schmir-at-bb:
Maybe I should explain one of the reasons I try to avoid this change  a bit more. Waiting for any of the nodes to be fullly synced once when starting may hide some bugs or make them harder to detect. The nodes may switch their state to syncing anytime and we need to be able to handle those cases.
Without waiting for a fully synced node, it's rather easy to test the bridge for the case where the node is syncing.

schmir-at-bb:
superseded by #487

schmir-at-bb:
> I could not find a good way to reduce redundancy in between wait_for_chain_synced_until_block and wait_for_chain_fully_synced_past_block without having one function that does two different things with a flag, since most of it is logging infos.

I think you should take a look at what I did in PR #487. You have to decompose things first to arrive at an elegant solution.

I've got a function that gets the status from the node and have a function that waits for a certain status to be reached, where the function that decides that the status we want is reached is passed in as a predicate function.

This is then used like in the following code:
https://github.com/trustlines-protocol/blockchain/blob/bce249fac2ee7ca5b33c460fc00b0d09ad8c6f30/tools/bridge/bridge/main.py#L409-L423

cducrest:
> > I could not find a good way to reduce redundancy in between wait_for_chain_synced_until_block and wait_for_chain_fully_synced_past_block without having one function that does two different things with a flag, since most of it is logging infos.
> 
> I think you should take a look at what I did in PR #487. You have to decompose things first to arrive at an elegant solution.
> 
> I've got a function that gets the status from the node and have a function that waits for a certain status to be reached, where the function that decides that the status we want is reached is passed in as a predicate function.
> 
> This is then used like in the following code:
> 
> https://github.com/trustlines-protocol/blockchain/blob/bce249fac2ee7ca5b33c460fc00b0d09ad8c6f30/tools/bridge/bridge/main.py#L409-L423

I think this mainly reduces redundancy because we no longer do two different things for the two nodes (wait for fully synced chain / wait for partially synced chain). But sure, your solution is more elegant and easier to adapt. Thanks for pointing it out.

PR #471, opened 2019-10-14
Add bumpversion config file for tools/bridge

schmir-at-bb:
We need to maintain the version of the bridge somehow. Using
setuptools_scm may not work perfectly, since we have multiple
independent components. Let's try bumpversion.

See https://github.com/c4urself/bump2version


schmir-at-bb:
> I have no clue about `bumpversion`. But nothing unexpected. I guess it is not possible to merge the file `tools/bridgeVERSION` and the configuration value in `bumpversion.cfg`?

bumpversion will maintain and keep the version number in sync it at least three files. this is a feature.
the `VERSION` file is meant to be viewed by a user or from a shell script.

weilbith:
That sounds awesome. I should get in touch with it. :hugs:

PR #475, opened 2019-10-15
Update readme file

jannikluhn:
Update readme file. Previously it only described the node setup, now it includes all the other components as well and should work with the latest version of the quickstart script and the other docker containers. Except that it doesn't (yet, see comments).

I've also added some more general info like (mostly guessed) system requirements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,81 +1,90 @@
 # Trustlines Blockchain
 
-- [Run Local Peer](#run-local-peer)
-  - [Quickstart](#quickstart)
-  - [System Time](#system-time)
-  - [With Docker](#with-docker)
-    - [Pre-Requisites](#pre-requisites)
-    - [Usage](#Usage)
-    - [Examples](#Examples)
-    - [Observer](#Observer)
-    - [Participant](#Participant)
-    - [Validator](#Validator)
-    - [Create New Account](#create-new-account)
-  - [Without Docker](#without-docker)
-    - [Pre-Requisites](#pre-requisites)
-    - [Creating An Account](#creating-an-account)
-    - [Setup For Users Using Only CLI](#setup-for-users-using-only-cli)
-    - [Setup For Validators Using Only CLI](#setup-for-validators-using-only-cli)
-- [Backups](#backups)
+- [The Trustlines Blockchain Infrastructure](#the-trustlines-blockchain-infrastructure)
+  - [System Requirements](#system-requirements)
+  - [Backups](#backups)
+  - [Setup With the Quickstart Script](#setup-with-the-quickstart-script)
+  - [Setup With Docker](#setup-with-docker)
+    - [Blockchain Node](#blockchain-node)
+    - [Netstats Client](#netstats-client)
+    - [Monitor](#monitor)
+    - [Bridge](#bridge)
+  - [Setup Without Docker](#setup-without-docker)
 - [Development](#development)
   - [Build Own Image](#build-own-image)
   - [Upload Image](#upload-image)
   - [Running Tests on Contracts](#running-tests-on-contracts)
 
-## Run Local Peer
+## The Trustlines Blockchain Infrastructure
 
-Please make sure you have access to a continuously running machine, if you like to participate as validator to the network.
+Nodes of the Trustlines Blockchain run various applications:
 
-### Quickstart
+- The node of the blockchain itself
+- The monitor that checks if validators act honestly
+- The bridge between Ethereum and the Trustlines Blockchain (only run by validators)
+- The netstats client to report the node state to `https://netstats.trustlines.foundation`(optional)
 
-To make starting a validator node for the Trustlines Network as quick as
-possible, the _quickstart_ script can be used. Simply download and run the
-script. The script will setup everything that is necessary. This includes the
-_Parity_ client, an automated update service and more. Therefore the script will
-interact with the user to request missing information. The script can be called
-multiple times without problems. It checks what is already setup and what is
-left. At least all processes get updated and restarted on each run. The _Parity_
-client will restart automatically on failures.
+There are multiple ways to set each of these up. The most straightforward one by far is via our interactive quickstart
+script. Finer control can be achieved by starting the components individually as Docker containers. Finally, it is also
+possible to avoid Docker altogether and run everything directly on the host machine.
+
+Before starting the installation process, please have a look at the system requirements and the note on backups.
+
+### System Requirements
+
+Based on the experiences we have had on our long-running testnet, we recommend at least 2GB of memory and 10GB of SSD
+storage.
+
+Validators should make sure their node has a high uptime: Otherwise, they miss out on potential revenue and harm the
+network by increasing average block intervals and time to finality.
+
+For block validation and creation, it is essential to make sure your host system has the correct time configuration. On
+most systems, this should be the case by default. You can check the settings with `timedatectl` (look for
+`"System clock synchronized: yes"`). For more information, see for example the corresponding
+[Ubuntu help page](https://help.ubuntu.com/lts/serverguide/NTP.html).
+
+For the quickstart and Docker installation modes, [Docker](https://docker.com) needs to be installed and configured.
+Please refer to the official documentation and make sure your user is added to the `docker` user group if you cannot
+access root permissions to run containers.
+
+### Backups
+
+For validators it is crucial to safely back up their private key.
+
+### Setup With the Quickstart Script
+
+The quickstart script will set up the blockchain node and the monitor as well as optionally the bridge and netstats
+clients. It allows importing a private key in order to act as a validator. In addition, it will start a
+[watchtower](https://hub.docker.com/r/containrrr/watchtower) to automatically update the containers when newer versions
+become available (e.g. for bug fixes or network forks).
+
+To fetch the most recent version of the quickstart script and run it, execute the following command on your machine:
 
 ```sh
 $ wget -O quickstart.sh https://github.com/trustlines-protocol/blockchain/raw/master/quickstart.sh && bash quickstart.sh
 ```
 
-Follow the instructions to insert your password. If you want to restart the node or want to make sure it runs on the most recent version, just
-rerun the script.
+The script is interactive and will ask you which components to set up. Once the setup is complete, the various
+components will run in the background in the form of Docker containers. Configuration and chain data can be found in
+the `trustlines` directory placed in the current working directory.
 
----
+Executing the script again is safe: No configuration will be overridden. This allows you to add components not
+configured in earlier runs and will restart all containers.
 
-### System Time
+### Setup With Docker
 
-Due to the way the block validation works and is synchronized, it is essential to make sure your host system has the correct time configuration. On recent Ubuntu systems, for example, this should already be the case. You can check the settings on such systems using `timedatectl`.
+A more explicit way of setting up the various components is starting the Docker containers manually as described here.
+To keep commands as concise as possible, only the most basic options are provided. You might want to set additional
+ones, e.g., container names or restart policies.
 
-On other operating systems you should check if your time is synchronized with an [NTP server](https://www.pool.ntp.org/).
+#### Blockchain Node
 
----
-
-### With Docker
-
-The following instructions explain how to start a local peer with the _Docker_ image. In fact it use a pre-configured [Parity
-Ethereum](https://www.parity.io/) client, combined with a set-up wrapper, to make connecting as easy as possible. The image is prepared to be
-used in three different scenarios: as observer, to participate and connect as validator.
-
-#### Pre-Requisites
-
-To make it work, a complete [Docker](https://www.docker.com/) environment is necessary to be installed on your system. Please take a look into
-the [official documentation](https://docs.docker.com/install/#general-availability) and use the instructions for your respective OS. Some
-niche operating systems are not mentioned there but still provide packages (e.g. [ArchLinux - docker](https://www.archlinux.org/packages/community/x86_64/docker/)).
-Make sure that your user is added to the `docker` user-group on _Unix_ systems, if you can not access root permissions to run containers.
-
-#### Usage
-
-To run the parity client for the Trustlines Network you first have to pull the image from
-[DockerHub](https://hub.docker.com/r/trustlines/tlbc-testnet). It does not matter in which directory your are working this step, cause it will
-be added to _Dockers_ very own database. Afterwards calling the help should give a first basic overview how to use.
+The blockchain image is a standard Parity client but with a custom configuration for the Trustlines Blockchain. It also
+accepts a few additional command line options as described in the help message:
 
 ```
-$ docker pull trustlines/tlbc-testnet
 $ docker run trustlines/tlbc-testnet --help
+
 
  NAME
    Parity Wrapper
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,280 +1,231 @@
...
-parity account new --chain trustlines-spec.json -d [path/to/node/foler]
+docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' laika-node
 ```
 
-For the rest of this documentation to become a validator we will assume you ran:
+Now, the netstats client can be started with
 
 ```sh
-parity account new --chain trustlines-spec.json -d validator_node
+$ docker run --name netstats-client --network network-laika -e trustlines/netstats-env -e RPC_HOST=<node IP address> -e RPC_PORT=8545 trustlines/netstats-client:master
```
jannikluhn: This command doesn't work. The netstats client connects to the node just fine via HTTP, but then apparently tries to make a websocket request which fails. Which is weird, because as far as I do pretty much the same as the docker compose file.

README.md
```diff
@@ -1,280 +1,231 @@
...
-For observing the Trustlines blockchain absolutely nothing has to be prepared for the client image. A simple pull and run is all that is needed.
-Since the loss of the block data when the container restarts would lead to a long startup procedure where the clients needs to sync up first, it is
-recommended to use a local volume bound in to save the data outside of the _Docker_ container. Furthermore the node should be able to explore the network and establish
-connections to other peers. Therefore the port `30300` has to be public available and mapped to the containers one.
+First of all, to conveniently allow other containers to connect to the node, create a docker network with
 
 ```sh
-$ mkdir ./database
-$ docker run -ti -v $(pwd)/database:/data -p 30300:30300 trustlines/tlbc-testnet
+$ docker network create network-laika
```
jannikluhn: Originally I just used the host network for everything to use as few commands as possible. Then I tried to add a network to fix the netstats bug which didn't work. Now I'm not sure what's better anymore.

README.md
```diff
@@ -1,280 +1,231 @@
...
+tokens on the Trustlines Blockchain. Non-validators should not run a bridge node.
 
-Independently of whether you created an account or not, you can start a node to sync with the trustlines chain with the following command:
+The bridge requires an Ethereum mainnet node which can be a light client. To start one, execute
 
 ```sh
-parity --chain trustlines-spec.json -d user_node --auto-update=none --no-download
+$ docker network create network-ethereum
+$ mkdir -p trustlines/databases/goerli
+$ docker run -d --name goerli-node --network network-ethereum -v $(pwd)/trustlines/databases/goerli:/data/database parity/parity:stable --light --no-download --auto-update none --chain goerli --db-path /data/database --base-path /data --no-hardware-wallets --jsonrpc-apis safe --jsonrpc-hosts all --jsonrpc-cors all --jsonrpc-port 8545 --no-ipc --no-secretstore --no-color
```
jannikluhn: This fails with a permission denied error:

```
Error upgrading parity data: CannotWriteVersionFile(Os { code: 13, kind: PermissionDenied, message: "Permission denied" })
```

If I don't use a volume I get

```
Error upgrading parity data: CannotCreateConfigPath(Os { code: 13, kind: PermissionDenied, message: "Permission denied" })
```

weilbith: Because of the shared volume. Have a look at [this comment](https://github.com/trustlines-protocol/blockchain/blob/0b72bd96d1f7987abfea2b1257743718453d7694/docker/docker-compose.yaml#L35)

weilbith: Format as the review comments above...

README.md
```diff
@@ -1,280 +1,231 @@
...
 ```
 
-#### Setup For Validators Using Only CLI
+Similar as for the Laika node, find its IP address using
 
-These are the steps to run a node as a validator, make sure you are a validator before following these steps.
+```sh
+docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' goerli-node
+```
```
weilbith: As before...

README.md
```diff
@@ -1,280 +1,231 @@
...
-in the volume bound to `/config` relative to [these instructions](#participant). When working with the quickstart script, the key is placed in
-`./trustlines/config/keys/Trustlines` relative to where you have run the script.
+- [Parity](https://wiki.parity.io/Parity-Ethereum)
+- [Netstats](https://github.com/trustlines-protocol/ethstats-client)
+- [Monitor](https://github.com/trustlines-protocol/tlbc-monitor)
+- [Bridge](https://github.com/trustlines-protocol/blockchain/tree/master/tools/bridge)
 
----
+For the Laika node, make sure it uses the correct [chain spec file](), that the right TCP and UDP ports are used
+(30300), and that the JSON RPC APIs `web3`, `eth`, and `net` are open.
```
weilbith:
```suggestion
(30300), and that the JSON RPC APIs `web3`, `eth`, and `net` are enabled.
```
To do not mix it with open the port. Just to make it foolproof regarding security.

README.md
```diff
@@ -1,280 +1,231 @@
...
-rerun the script.
-
----
+Before starting the installation process, please have a look at the system requirements and the note on backups.
 
-### System Time
+### System Requirements
 
-Due to the way the block validation works and is synchronized, it is essential to make sure your host system has the correct time configuration. On recent Ubuntu systems, for example, this should already be the case. You can check the settings on such systems using `timedatectl`.
+Based on the experiences we have had on our long-running testnet, we recommend at least 2GB of memory and 10GB of SSD
```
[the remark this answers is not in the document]
jannikluhn: True, I guess it's something we should double check, but for now it's our best estimate. I suspect we're not too far off though.

README.md
```diff
@@ -1,280 +1,231 @@
...
+Validators should make sure their node has a high uptime: Otherwise, they miss out on potential revenue and harm the
+network by increasing average block times.
 
----
+For block validation and creation, it is essential to make sure your host system has the correct time configuration. On
+most systems, this should be the case by default. You can check the settings with `timedatectl` (look for
+`"System clock synchronized: yes"`).
 
-### With Docker
+For the quickstart and docker installation modes, [docker](https://docker.com) needs to be installed and configured.
```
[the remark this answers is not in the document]
jannikluhn: I'd rather not link to this directly as it contains a changing version number. But users should be capable to find it on their own, so I think it's fine.

README.md
```diff
@@ -1,280 +1,231 @@
...
+Please refer to the official documentation and make sure your user is added to the `docker` user group if you cannot
+access root permissions to run containers.
 
-The following instructions explain how to start a local peer with the _Docker_ image. In fact it use a pre-configured [Parity
-Ethereum](https://www.parity.io/) client, combined with a set-up wrapper, to make connecting as easy as possible. The image is prepared to be
-used in three different scenarios: as observer, to participate and connect as validator.
+### Backups
 
-#### Pre-Requisites
+For validators it is crucial to safely back up their private key.
```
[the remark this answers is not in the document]
jannikluhn: I moved it up explicitly to make it more visible and because the reading paths diverge depending on the installation method, so few people will read until the end.

Since validators participated in the auction with the same key already, they should have backups at this point and most people won't have to use `scp`.

README.md
jannikluhn: I'd like to hold on with naming changes here until we've sorted them out globally. Same with testnet vs mainnet. We'll have to make another run through at this point anyways..

README.md
```diff
@@ -1,280 +1,231 @@
...
+$ echo "<passphrase_for_keystore_file>" > trustlines/config/pass.pwd
+$ docker run -d --name laika-node --network network-laika -v $(pwd)/trustlines/databases:/data -v $(pwd)/trustlines/config:/config/custom -v $(pwd)/trustlines/enode:/config/network -v $(pwd)/trustlines/shared:/shared/ -p 30300:30300 -p 30300:30300/udp trustlines/tlbc-testnet --role validator --address MY_ADDRESS
+```
 
-#### Pre-Requisites
+#### Netstats Client
 
-- parity version 2.0.9
-- the file: "trustlines-spec.json"
+The netstats client reports the state of your node to [https://laikanetstats.trustlines.foundation/]() that gives a
```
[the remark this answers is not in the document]
jannikluhn: see above note on test vs main net

README.md
```diff
@@ -1,280 +1,231 @@
...
-cd trustlines-chain
+HIDE_VALIDATOR_STATUS=false
 ```
 
-You can then create an account with the command:
+The netstats client has to connect to the blockchain node. Therefore, we have to find out the node's IP address with
 
 ```sh
-parity account new --chain trustlines-spec.json -d [path/to/node/foler]
+docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' laika-node
```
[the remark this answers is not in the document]
jannikluhn: this was before the bridge config fixes, so this wouldn't have worked. Now it should, yes

README.md
```diff
@@ -1,280 +1,231 @@
...
+$ mkdir -p trustlines/monitor/reports trustlines/monitor/state
+$ docker run -d --name tlbc-monitor --network network-laika -v $(pwd)/trustlines/shared:/config -v $(pwd)/trustlines/monitor/state:/state -v $(pwd)/trustlines/monitor/reports:/reports trustlines/tlbc-monitor -c /config/trustlines-spec.json -r /reports -d /state -u http://<node IP address>:8545
 ```
 
-#### Setup For Users Using Only CLI
+#### Bridge
 
-These are the steps to run a node as a simple user.
+Validators of the Trustlines Blockchain have to run the bridge that converts TLN tokens on the Ethereum chain to TLC
+tokens on the Trustlines Blockchain. Non-validators should not run a bridge node.
```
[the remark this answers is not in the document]
jannikluhn: To me this is a purely technical distinction that shouldn't matter to the user. Ether I think is worse because it refers to the Ethereum chain.

README.md
jannikluhn: I've seen this comment, but I'm not sure if this is the issue. That's why I tried not using a volume at all and it still failed.

Also, if it is the issue, how can we fix it? We don't have a compose file here.

README.md
weilbith: Okay, that sounds reasonable. :+1:

README.md
weilbith: > That's why I tried not using a volume at all and it still failed.

So just as a standard node without a validator key?

> Also, if it is the issue, how can we fix it? We don't have a compose file here.
1. Compose files can't add options to a container that _Docker_ does not support. So you can just use `docker run --user...` herefore.
2. You can also just use (a) different volume(s) for this node. :man_shrugging:

README.md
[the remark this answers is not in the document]
weilbith: Okay. If I remember it correctly @schmir-at-bb was working on issue about resources anyways. Maybe he can then adjust the recommendation as well.
Currently we already allow the mainnet node, bridge client and monitor to use together up to 2GB. The tlbc node and netstats client (I don't count watchtower) have no limits so far.

README.md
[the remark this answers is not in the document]
weilbith: That is true.

README.md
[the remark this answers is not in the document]
weilbith: Okay, I'm fine with that opinion. The key is the only thing that actually matters.

README.md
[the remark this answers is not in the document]
weilbith: So the reason is? Because this way is cumbersome and less obvious. In my opinion. Anyways it is still a working solution.

weilbith:
@jannikluhn what is the working status? Could be many of these conversations closed?

README.md
jannikluhn: Now it seems to not even be able to make the HTTP request.

README.md
jannikluhn: I tried (2), but it doesn't work. I get the same error message as above.

I've added the `--user root` option, but I don't think this is an acceptable solution. Root permissions should not be necessary to run a node.

jannikluhn:
I've updated everything as far as I could. Some commands still don't work though:

- netstats can't connect to node
- goerli node doesn't run without root
- bridge doesn't run, because of the config parsing issue which is already fixed but didn't make it to docker hub yet.

README.md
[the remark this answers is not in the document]
schmir-at-bb: the live test servers had 4GB of main memory available. I would rather recommend 4GB of main memory to be on the safe side. If we like to recommend the 2GB of memory, we should at least test it.
 
the 10GB of SSD storage are too small. I've got a testnet node running that's already using 9.2GB of storage.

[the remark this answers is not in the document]
weilbith: Yes. I think the old calculations of Michael are not much worth anymore.
Thanks for your input Ralf!

weilbith:
@jannikluhn I got your netstats problem. You create an environment file but don't use it. So you get denied by the netstats server not the blockchain node. In contrast to `docker-compose` does `docker run` not source the `.env` file automatically. Anyways you recommend a file like for the quickstart setup. So you must reference this in the command like `docker run --env-file ./trustlines/netstats-env ... trustlines/netstats-client`. Try this and it should work then.

jannikluhn:
Thanks you're right, I did reference it, but with `-e` instead of `--env-file` which is the wrong option. The other fix necessary was to change `RPC_HOST=http://laika-node` to `RPC_HOST=laika-node`. Now it works. Yay.

jannikluhn:
Updated the hardware suggestion and this should be ready from my side now.

weilbith:
> bridge doesn't run, because of the config parsing issue which is already fixed but didn't make it to docker hub yet.

@jannikluhn can you link the related issue which will upload all the `*/release` tags to the repos?

jannikluhn:
Not sure what you mean, here's the PR that I was referring to: https://github.com/trustlines-protocol/blockchain/pull/477

This one is about release tags, but I don't see how it is related: https://github.com/trustlines-protocol/blockchain/pull/488

README.md
```diff
@@ -1,280 +1,261 @@
...
+```
 
-### With Docker
+The script is interactive and will ask you which components to set up. Once the
+setup is complete, the various components will run in the background in the form
+of Docker containers. Configuration and chain data can be found in the
+`trustlines` directory placed in the current working directory. It is possible
+to customize the own setup by editing those configuration files. This goes for
+the configuration of the different components, as well as the composition of the
+Docker containers. If an optional component has not been setup on an earlier
```
cducrest:
```suggestion
Docker containers. If an optional component has not been set up on an earlier
```

README.md
```diff
@@ -1,280 +1,261 @@
...
-#### Pre-Requisites
+### Setup With Docker
 
-To make it work, a complete [Docker](https://www.docker.com/) environment is necessary to be installed on your system. Please take a look into
-the [official documentation](https://docs.docker.com/install/#general-availability) and use the instructions for your respective OS. Some
-niche operating systems are not mentioned there but still provide packages (e.g. [ArchLinux - docker](https://www.archlinux.org/packages/community/x86_64/docker/)).
-Make sure that your user is added to the `docker` user-group on _Unix_ systems, if you can not access root permissions to run containers.
+A more explicit way of setting up the various components is starting the Docker containers manually as described here.
+To keep commands as concise as possible, only the most basic options are provided. You might want to set additional
+ones, e.g., container names or restart policies. Notice that we intent to
```
cducrest:
```suggestion
ones, e.g. container names or restart policies. Notice that we intent to
```

README.md
```diff
@@ -1,280 +1,261 @@
...
+rebuild to setup of the quickstart script within the following sections. It is
```
cducrest: `Notice that we intent to rebuild to setup of the quickstart script within the following sections.`
I do not understand this sentence, if you are ok with removing it, please do.

README.md
```diff
@@ -1,280 +1,261 @@
...
+a complete setup you could also first use the quickstart and adjust the setup
+afterwards, as it is fully customizable.
 
-#### Usage
+#### Blockchain Node
 
-To run the parity client for the Trustlines Network you first have to pull the image from
-[DockerHub](https://hub.docker.com/r/trustlines/tlbc-testnet). It does not matter in which directory your are working this step, cause it will
-be added to _Dockers_ very own database. Afterwards calling the help should give a first basic overview how to use.
+The blockchain image is a standard Parity client but with a custom configuration for the Trustlines Blockchain. It also
```
cducrest:
```suggestion
The blockchain image is a standard Parity client with a custom configuration for the Trustlines Blockchain. It also
```

README.md
```diff
@@ -1,280 +1,261 @@
...
----
-
-### Without Docker
+#### Netstats Client
 
-This section explains how to start a local peer without using the _Docker_ image.
+The netstats client reports the state of your node to the
+[Laika netstats page](https://laikanetstats.trustlines.foundation/) that gives a rough overview of the current network state.
+It is a fully optional component which helps the community, at the cost leaks of some leakage of
+private information.
```
cducrest: In correct English: `It is a fully optional component which helps the community at the cost of leakage of some private information.`
I would simplify it as: `It is an optional component which helps the community by providing information on your running node to a central server.`

README.md
weilbith: Copy of our _RocketChat_ conversation:
Hmm. My problem was that the descriptions by Jannik where quite “complicted” without a hint why they are like that. So he reconstructs the `trustlines` directory structure and all that. So in fact he is doing what the quickstart script is doing, but the manual way.

README.md
```diff
@@ -1,280 +1,261 @@
...
+bridge_contract_address = "0x854dF872BB8bfECafFB1077FCfd7aa0B7C838A60"
+event_fetch_start_block_number = 4153205
+
+[validator_private_key]
+keystore_path = "/config/keys/Trustlines/<keystore_filename.json>"
+keystore_password_path = "/config/pass.pwd"
 ```
 
-Otherwise, you need to adjust some parameters. You need to provide the path to your private key in "--keys-path=[path]", the path to your password protecting this key in "--password=[path]" as well as your address in "--author=[address]" and "--engine-signer=[address]"
+Note that they keystore path is not an actual path on the host machine, but rather in the bridge container we will
```
cducrest:
```suggestion
Note that the keystore path is not an actual path on the host machine, but rather in the bridge container we will
```

cducrest:
```suggestion
Note that the keystore path is not an actual path on the host machine, but rather in the bridge container we will
```

PR #478, opened 2019-10-15
Update contract-deploy-tools version to 0.6.1

weilbith:
Related to: trustlines-network/project#703


weilbith:
The end2end tests fail, because the used Parity version is to old and is missing the petersburg EIPs. They are introduced per default to the `dev` chain on `Parity v2.6.4`.

PR #486, opened 2019-10-16
Copy docker-compose file to host during quickstart

berndbohmeier:
This copies the docker-compose file to the host during the quickstart
script. This allows to later change paramters to containers and also the
used images.
This makes the quickstart script configurable based on the
docker-compose file used, you can change the name of the install folder,
and you can change the project name used by docker-compose.
To not just simply override user changes, we give him the chance to show
a diff.

### DIFF
--- a/tools/quickstart/quickstart/docker.py
+++ b/tools/quickstart/quickstart/docker.py
@@ -1,4 +1,5 @@
 import difflib
+import filecmp
 import os
 import shutil
 import subprocess
@@ -19,10 +20,13 @@ from quickstart.validator_account import get_validator_address
 # List of docker container names to stop and remove on startup in addition to the ones defined in
 # the docker compose file (for backward compatibility)
 LEGACY_CONTAINER_NAMES = ["watchtower-testnet", "trustlines-testnet"]
+DOCKER_COMPOSE_FILE_NAME = "docker-compose.yaml"
 
 
 def setup_interactivaly(base_dir, docker_compose_file):
-    if does_docker_compose_file_exist(base_dir):
+    if does_docker_compose_file_exist(base_dir) and not filecmp.cmp(
+        os.path.join(base_dir, DOCKER_COMPOSE_FILE_NAME), docker_compose_file
+    ):
         while True:
             choice = click.prompt(
                 fill(
@@ -59,7 +63,7 @@ def update_and_start(*, base_dir, host_base_dir, project_name) -> None:
             "\n"
             + fill(
                 "Expecting a docker-compose configuration file at the current directory "
-                "with a standard name. ('docker-compose.yaml' or 'docker-compose.yml')"
+                f"with name '{DOCKER_COMPOSE_FILE_NAME}'"
             )
         )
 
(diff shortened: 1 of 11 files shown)


tools/quickstart/quickstart/docker.py
```diff
@@ -19,10 +21,40 @@
 LEGACY_CONTAINER_NAMES = ["watchtower-testnet", "trustlines-testnet"]
 
 
-def update_and_start(host_base_dir: str) -> None:
-    if not os.path.isfile("docker-compose.yaml") and not os.path.isfile(
-        "docker-compose.yml"
-    ):
+def setup_interactivaly(base_dir, docker_compose_file):
+    if does_docker_compose_file_exist(base_dir):
+        while True:
```
jannikluhn: How about we check if the two files are equal first (e.g. with `filecmp.cmp`) and if so just do nothing? Then we wouldn't have to bother the user in the normal case that they didn't touch the docker file.

tools/quickstart/quickstart/docker.py
```diff
@@ -32,29 +64,43 @@ def update_and_start(host_base_dir: str) -> None:
...
         click.echo("\nNode will run as a non-validator")
 
+    with open(os.path.join(base_dir, ".env"), mode="w") as env_file:
+        env_file.writelines(
+            f"{key}={value}\n" for (key, value) in env_variables.items()
+        )
+
+    runtime_env_variables = {
+        **os.environ,
+        **env_variables,
```
jannikluhn: Any reason why we `env_variables` are both written to `env_file` and used as runtime vars?

tools/quickstart/quickstart/docker.py
```diff
@@ -110,19 +156,68 @@ def update_and_start(host_base_dir: str) -> None:
...
         docker_service_names.append("bridge-client")
         docker_service_names.append("mainnet-node")
 
     return docker_service_names
+
+
+def does_docker_compose_file_exist(base_dir):
+    return os.path.isfile(
+        os.path.join(base_dir, "docker-compose.yaml")
+    ) or os.path.isfile(os.path.join(base_dir, "docker-compose.yml"))
```
jannikluhn: Maybe slightly more readable:

```Py
filenames = [os.path.join(base_dir, "docker-compose." + extension) for extension in ["yml", "yaml"]]
return any(os.path.isfile(filename) for filename in filenames)
```

berndbohmeier:
@jannikluhn Is that solving your configuration issue? This way you can not really configure before running the quickstart, but after it is setup, you can just change the docker-compose file.

tools/quickstart/quickstart/docker.py
berndbohmeier: I can do that, but we would still bother him in case we changed something, but he did still not touch it. I thought about having a list of old file hashes, but decided that it is not forth the work at the moment. 
I can add the same file check for now.

tools/quickstart/quickstart/docker.py
berndbohmeier: Yes, the file is for later running it with the docker compose file, the runtime vars for the setup during the quickstart. Because we are writing it to a different folder, the env file is not so easy accessible during quickstart, so I decided the easiest is just to add it to runtime as well. The other option would be to switch to the right folder before executing the docker compose commands, do you think that is better?

jannikluhn: I see, makes sense. I'm ok with the current solution then.

jannikluhn:
Yes, this would have solved the problems I had.

tools/quickstart/quickstart/docker.py
berndbohmeier: I removed handling both extensions, as it would have not worked properly anyways with the new copying.

Issue #473, opened 2019-10-14
Fix: Wait for home chain to be fully synced


berndbohmeier:
When restarting the bridge it sent out a lot of confirmation for already completed transfers. This seems to be because we do not properly wait for the home node to be synced.

schmir-at-bb:
We should not wait at startup and instead ask for the syncing status in the event_fetcher (unless there is some other kind of bug, which I don't think is the case).

berndbohmeier:
No, I think it is just that we do not wait properly for it to be synced, and we already saw then at other places that the getLogs endpoint might just return no events.

berndbohmeier:
So waiting in the fetcher should be the right thing, yes.

schmir-at-bb:
> No, I think it is just that we do not wait properly for it to be synced, and we already saw then at other places that the getLogs endpoint might just return no events.

But that would be a different issue that you want to solve. If a light client returns no events if it's not fully synced, we can take a look at it in another issue.
 
> So waiting in the fetcher should be the right thing, yes.

No, *waiting* is the wrong thing to do. Just ask for the syncing status before putting the FetcherReachedHeadEvent on the queue is enough.  (We can either not put the event on the queue if the node is still syncing or put an additional event on the queue with the syncing status, or pass the syncing status inside the event)

That way we will still be able to process events, but we will still not send out confirmations.

PR #487, opened 2019-10-16
Wait for fully synced nodes

schmir-at-bb:
This is based on #485, please only review the last two commits.

Wait for home and foreign node to be fully synced

We decided that it might be safer to wait until both nodes are fully
synced.

There are two additional advantages:

1. We do not run into an error when we ask for a contracts code in
order to do our sanity checks.
(see https://github.com/trustlines-protocol/blockchain/issues/452)

2. We prevent the problem with warp-syncing which has a gap of blocks
for which parity does only deliver empty results when we ask for
events.
(see https://github.com/trustlines-protocol/blockchain/issues/473)

The code has been refactored and cleaned up a bit. We now use
bridge.node_status module.

This supersedes PR #459

### DIFF
--- a/tools/bridge/bridge/node_status.py
+++ b/tools/bridge/bridge/node_status.py
@@ -1,0 +1,72 @@
+import logging
+from typing import Any, Dict, List
+
+import attr
+import gevent
+import tenacity
+
+logger = logging.getLogger(__name__)
+
+
+@attr.s(auto_attribs=True)
+class NodeStatus:
+    is_parity: bool
+    is_light_node: bool
+    is_syncing: bool
+    syncing_map: Dict
+    block_number: int
+    latest_synced_block: int
+    block_gap: List[int]
+    client_version: Any = attr.ib(repr=False)
+
+
+def get_node_status(w3):
+    client_version = w3.clientVersion
+    is_parity = client_version.startswith("Parity")
+    block_number = w3.eth.blockNumber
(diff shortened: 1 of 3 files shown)


berndbohmeier:
`Make SIGTERM/SIGINT handling work during startup phase` is already in master, please
bring this Pr up to sync for easier review.

schmir-at-bb:
@jannikluhn ping

tools/bridge/bridge/node_status.py
```diff
@@ -0,0 +1,91 @@
...
+    node_kind = w3.manager.request_blocking("parity_nodeKind", [])
+
+    if chain_status.blockGap is not None:
+        block_gap = [int(x, 16) for x in chain_status.blockGap]
+    else:
+        block_gap = None
+
+    is_light_node = node_kind.capability == "light"
+
+    if block_gap and not is_light_node:  # warp mode syncing
```
berndbohmeier: Why do you not just have `if block_gap`?

schmir-at-bb: the light mode node always has a block gap over the full range of blocks.

berndbohmeier: ah ok

schmir-at-bb:
> Using that also while the bridge is running will come in a different PR I assume

yes

PR #488, opened 2019-10-17
Add release tags to all images

jannikluhn:
Doesn't work yet because some of the tags don't exist yet.

Also added on to the tlbc-testnet image for consistency (as far as I can tell it shouldn't make a difference, at least latest == release right now at docker hub).


berndbohmeier:
Is there the work in progress to get all the release tags?

jannikluhn:
No, not that I know of.

berndbohmeier:
Outdated

PR #495, opened 2019-10-21
Add default config Laika to quickstart script

berndbohmeier:
The quickstart script now has a default option for the Laika setup. This
will be also used later to have another default setup for trustnet.
This works because we now bundle the docker-compose file and the bridge
config.

quickstart.sh is still waiting for this all to be finished.


tools/quickstart/quickstart/cli.py
```diff
@@ -1,52 +1,146 @@
...
     ),
     type=click.Path(),  # Can not check for exist, because path is on host
     default=None,
 )
 @click.option(
     "--docker-compose-file",
-    help="path to the docker compose file to use",
-    type=click.Path(exists=True, dir_okay=False),
-    default="docker-compose.yaml",
+    help="Path to the docker compose file to use, or laika for the default laika docker compose file.",
```
weilbith:
```suggestion
    help=f"Path to the docker compose file to use, or '{LAIKA}' for the default laika docker compose file.",
```

weilbith: I stumbled other this while reading. Even having seen the code above before. So I think adding these quotes make it more obvious that it is a specific keyword.
Moreover I would like to avoid stupid mistakes and replace the literal `laika` with `{LAIKA}`. Because that's what used to determine the keyword.
Same goes for the following `--bridge-config` option.

tools/quickstart/quickstart/cli.py
```diff
@@ -1,52 +1,146 @@
...
+            value = self.default_path_lookups[value]()
+
+        return super().convert(value, param, ctx)
 
 
 @click.command()
+@click.option(
+    "--config",
+    "-c",
+    help="Config to use. [Default: Laika]",
```
weilbith: This help is pretty short and not useful. Which of all the configs? It turns out that it defines presets for all following options. Maybe this should be mentioned here. Else I would think: If I set Laika here, I go for setting it at the other options as well. But then I get the error message that I can not combine them. Thereby the usage is not obvious.

weilbith: After all, I'm thinking if a command group is not more clear to use. So we have one, where you can define a preset and one where everything must be manually set. This would also make the top body of this function more understandable. Moreover the whole default path and getter can be removed. This would also allow to simply add function descriptions which are then parsed as help text by click. The user clearly see that he can either go the "automatic" way or the manual one.
Share you opinion and feel free to simply close this.

berndbohmeier: what would the command group be called?

weilbith: Actually a group is not necessary. Both commands do not share their root functionality. So it could be simply to different commands. The only point for a group would the shared option `--project-name`. Not sure if it is worth to establish a group for that, since it requires to maintain the context object. We would end up with something ugly like `quickstart --project-name trustlines4ever manual --bridge-config ...`.

So a proposal with pretty literal command names would be something like:
`quickstart use-defaults <chain-name>` ~> have a switch-case for the `chain-name` argument
`quickstart ask-for-settings <options>` ~> all options are simply required and must not be checked

berndbohmeier: The command names are pretty bad, so with this names I would leave it like we have it. 
Also remember the default is that you just do `quickstart`, so that also has to work.

berndbohmeier: we could have laika as a command 
so `quickstart` that can take options
or `quickstart laika` that does not take options
`quickstart` without options will fall back to quickstart laika

weilbith: That looks actually neat. And then later add `quickstart trustnet` (`tlbc`...) ?

berndbohmeier: yep

tools/quickstart/quickstart/cli.py
berndbohmeier: done

tools/quickstart/quickstart/cli.py
berndbohmeier: done

berndbohmeier:
@weilbith please have another look.

berndbohmeier:
Currently they are just copy change, but beside of the text they will differ.

PR #496, opened 2019-10-21
Write a smoketest for the blockchain docker image

berndbohmeier:
This will only do a basic check if the image did not crash directly
after startup.


berndbohmeier:
As you can see this smoketest now successfully demonstrates the currently broken docker file.

PR #497, opened 2019-10-22
Use node_status in event_fetcher

schmir-at-bb:
This makes the event fetcher work with warp-mode enabled nodes.

See https://github.com/trustlines-protocol/blockchain/issues/473


schmir-at-bb:
we should still turn if off to protect our users. I fail to see a reason, why we need warp-mode and it does have serious drawbacks.

PR #498, opened 2019-10-22
Add TLBC config

berndbohmeier:
This adds the configuration for Trustnet. Beside that we wait on if we should now use the name Trustnet, I have some more questions where I would like to get some input:

- We have now two Docker files for both chains which are pretty much the same. Should I try to spend time to share the code? (I dont think so). The only way I see would be to have a third Dockerfile that builds an intermediary image that both will build on. 

- We now have two docker-compose files that are pretty much the same. Should I spend time to share the code? The only way I know to use it would be with the override mechanism, this would need me to copy out several docker-compose files in the quickstart. Or we would need an own mechanism to create the docker-compose file from a template. (I am not sure here, because both would make the quickstart more complicated. )

- We used the names trustlines-node and mainnet-node before in the docker-compose file. I think this is a bad naming, because it is not a mainnet node on görli. I changed it to foreign-node and home-node, which is right from a bridge perspective, but those words might be not so familiar. What do you like better?


weilbith:
**Dockerfile:**
The differences between both images should be pretty small. Therfore I think we could introduce `ARG`s for them. Set their defaults to the `trustnet` version and add a `README` to this folder to document how to set the args for building the laika version.

**docker-compose**:
I'm not against having always two (in total three) `docker-compose.yaml` which get merged. It's easier for us to maintain. A user who wants to edit his setup should be also able to handle two merged files. 

**names**:
I think `home-node` and `foreign-node` are pretty simple and easy to understand. At least for me after all this time with the bridge. But at least running it as non-validator ends up with a single container named `home-node`. Could be confusing. I'm against using `ethereum` node like @jannik suggested. Else I'm also open for good suggestions.

PR #506, opened 2019-10-25
Bridge/release

schmir-at-bb:
Release 0.2.0


schmir-at-bb:
Trying with a different branch name.

PR #507, opened 2019-10-25
Bridge release 0.2.0

schmir-at-bb:


schmir-at-bb:
that one picked up the existing circleci build failures

PR #523, opened 2019-10-29
Feature/use rpc for foreign node

cducrest:
closes https://github.com/trustlines-network/project/issues/722

I may need to find a better solution than checking `is_bridge_using_default_json_rpc` to know whether we should start the foreign node or not.

What if the user changed the rpc of his foreign node locally and in the bridge config, this will no longer work. But do we consider that it is the user's fault for sabotaging its config?

Otherwise, some value needs to be passed from bridge's `setup_interactively` to docker's `update_and_start` to indicate whether to start the foreign node or not.

### DIFF
--- a/tools/quickstart/quickstart/bridge.py
+++ b/tools/quickstart/quickstart/bridge.py
@@ -5,6 +5,9 @@ from pathlib import Path
 from textwrap import fill
 
 import click
+import toml
+from marshmallow.exceptions import ValidationError
+from marshmallow.validate import URL
 
 from quickstart.constants import BRIDGE_CONFIG_FILE_EXTERNAL, BRIDGE_DOCUMENTATION_URL
 from quickstart.utils import (
@@ -34,7 +37,7 @@ def setup_interactively(base_dir, bridge_config_file) -> None:
         return
 
     if not is_validator_account_prepared(base_dir):
-        click.echo("\nNo bridge node will be set up as running as a non-validator.")
+        click.echo("\nNo bridge node will be set up when running as a non-validator.")
         return
 
     click.echo(
@@ -56,6 +59,23 @@ def setup_interactively(base_dir, bridge_config_file) -> None:
         # Necessary to make docker-compose not complain about it.
         Path(user_file).touch()
         return
+
+    choice = click.prompt(
+        fill(
+            "The bridge client requires access to a main chain node."
+            "Do you want to run a main chain node (1) or provide a JSON RPC url (2)?"
+        )
+        + "\n",
+        type=click.Choice(("1", "2")),
+        show_choices=False,
+    )
+    if choice == "1":
+        pass
+    elif choice == "2":
+        url = read_json_rpc_url()
+        change_json_rpc_url(bridge_config_file, url)
+    else:
+        assert False, "unreachable"
 
     copy_default_bridge_config(base_dir, bridge_config_file)
 
@@ -95,3 +115,24 @@ def _show_file_override_dialog(base_dir, bridge_config_file):
             )
         else:
             assert False, "unreachable"
+
+
+def read_json_rpc_url():
+    url = click.prompt(
+        "JSON RPC url"
+    )
+    try:
+        validator = URL()
+        validator(url)
+        return url
+    except ValidationError:
+        url = click.prompt(f"The given url seems invalid {url}, please correct it or enter the same url if valid")
+        return url
+
+
+def change_json_rpc_url(config_file, url):
+    config = toml.load(config_file)
+    config['foreign_chain']['rpc_url'] = url
+
+    with open(config_file, mode="w") as file:
+        toml.dump(config, file)
--- a/tools/quickstart/quickstart/bridge.py
+++ b/tools/quickstart/quickstart/bridge.py
@@ -6,8 +6,8 @@ from textwrap import fill
 
 import click
 import toml
-from marshmallow.exceptions import ValidationError
-from marshmallow.validate import URL
+from requests.exceptions import ConnectionError
+from web3 import HTTPProvider, Web3
 
 from quickstart.constants import BRIDGE_CONFIG_FILE_EXTERNAL, BRIDGE_DOCUMENTATION_URL
 from quickstart.utils import (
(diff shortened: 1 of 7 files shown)


tools/quickstart/quickstart/bridge.py
```diff
@@ -57,6 +59,21 @@ def setup_interactively(base_dir, bridge_config_file) -> None:
         Path(user_file).touch()
         return
 
+    choice = click.prompt(
+        fill(
+            "The bridge client requires access to a main chain node."
```
berndbohmeier: I would rather ask something like Do you have a main chain node running and provide the json rpc, or run a mainchain light node.

berndbohmeier: Also is main chain node the right term? Should we not rather say Ethereum Main Chain or Görli depending on which setup. Or Foreign chain node.

tools/quickstart/quickstart/bridge.py
```diff
@@ -57,6 +59,21 @@ def setup_interactively(base_dir, bridge_config_file) -> None:
...
+            "Do you want to run a main chain node (1) or provide a JSON RPC url (2)?"
+        )
+        + "\n",
+        type=click.Choice(("1", "2")),
+        show_choices=False,
+    )
+    if choice == "2":
+        url = read_json_rpc_url()
+        change_json_rpc_url(bridge_config_file, url)
+    elif choice != "1":
```
berndbohmeier: what is this?

tools/quickstart/quickstart/bridge.py
```diff
@@ -95,3 +112,31 @@ def _show_file_override_dialog(base_dir, bridge_config_file):
...
             assert False, "unreachable"
+
+
+def read_json_rpc_url():
+    url = click.prompt("JSON RPC url")
+    if verify_url(url):
+        return url
+    else:
+        url = click.prompt(
+            f"The given url seems invalid {url}, please correct it or enter the same url if valid"
```
berndbohmeier: Thats is weird, just ask the user again? why should it succeed the second time?

tools/quickstart/quickstart/bridge.py
```diff
@@ -95,3 +112,31 @@ def _show_file_override_dialog(base_dir, bridge_config_file):
...
+        )
+        return url
+
+
+def verify_url(url):
```
berndbohmeier: Please write a test for this, as we fucked it up already.

tools/quickstart/quickstart/docker.py
```diff
@@ -189,7 +190,8 @@ def get_optional_docker_service_names(base_dir) -> List[str]:
 
     if is_bridge_prepared(base_dir=base_dir):
         docker_service_names.append("bridge-client")
-        docker_service_names.append("foreign-node")
+        if is_bridge_using_default_json_rpc(base_dir):
```
berndbohmeier: Is there no better way? 
What if someone choose to run there node under the same url?

tools/quickstart/quickstart/bridge.py
cducrest: That is how I understood the requirement: we try to validate the URL, if it looks wrong we warn the user but let him proceed anyway. 

The way to let him proceed could be to ask him whether the URL is correct and let him change it if there was just a mistake, or to directly let him retype it. I chose the latter.

tools/quickstart/quickstart/bridge.py
cducrest: Would we rather not use marshmallow then? This way of doing it is what I understood from Ralf's suggestions, but I do not really know why it is better than trusting marshmallow directly for verification.

tools/quickstart/quickstart/docker.py
cducrest: The other way would be to ask whether they want to use their own URL outside of `bridge.setup_interactively` or return this information from `bridge.setup_interactively` so that we can give it to `docker.update_and_start` but I do not know if it is better as we still want to ask whether they want to use their own URL after we ask whether they want to set up the bridge.

tools/quickstart/quickstart/bridge.py
cducrest: I guess safety check, I took it from: https://github.com/trustlines-protocol/blockchain/blob/master/tools/quickstart/quickstart/validator_account.py#L72
I don't mind removing it.

tools/quickstart/quickstart/bridge.py
schmir-at-bb: It's a new dependency for the quicktools package. Other than that there's no good reason not to use it.

tools/quickstart/quickstart/bridge.py
berndbohmeier: Either then write it like that where you took it from or remove it. 
The reason there was to list all possible values and have an else to catch if you missed one.

tools/quickstart/quickstart/bridge.py
berndbohmeier: I thought this was about making the test call to the rpc endpoint, which is completely missing. 
I dont get you workflow, so if I misstyped it a second time I am doomed?

tools/quickstart/quickstart/bridge.py
berndbohmeier: What has writing a test to do with using marshmallow

cducrest: I can understand that we do not trust this verify_url function I wrote and require tests for it. I would trust marshmallow verification and not write tests. But I will just write tests.

tools/quickstart/quickstart/bridge.py
cducrest: changed.

tools/quickstart/quickstart/bridge.py
```diff
@@ -57,6 +59,24 @@ def setup_interactively(base_dir, bridge_config_file) -> None:
...
+            "or do you wish to run a foreign chain light node(2)?"
+        )
+        + "\n",
+        type=click.Choice(("1", "2")),
+        show_choices=False,
+    )
+    if choice == "1":
+        # Nothing to do
+        pass
+    if choice == "2":
```
berndbohmeier: elif

tools/quickstart/quickstart/bridge.py
```diff
@@ -57,7 +60,23 @@ def setup_interactively(base_dir, bridge_config_file) -> None:
...
         return
 
-    copy_default_bridge_config(base_dir, bridge_config_file)
+    choice = click.prompt(
+        fill(
+            "Do you have a main chain node running and provide the JSON rpc url (1), "
+            "or do you wish to run a foreign chain light node (2)?"
+        )
+        + "\n",
+        type=click.Choice(("1", "2")),
```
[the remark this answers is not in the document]
berndbohmeier: but the default stuff will not work then, if I just hit enter without inputing something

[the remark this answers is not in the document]
cducrest: Yes, do you want it?

Issue #464, opened 2019-10-14
Push and use bridge release tag in docker-compose file


berndbohmeier:

weilbith:
Status:
- `trustlines/tlbc` repository does not exist and is actually named `trustlines/tlbc-node` as I know
- `trustlines/bridge-release` repository does not exist and is probably a typo and `release` means the tag (as for the `laika` composition)
- all other repositories and their `release` tag exist

The composition for `laika` works well as it is (at least for `docker-compose pull`). I'll go to open a small PR adjusting both issues.

weilbith:
I just realized that the `trustlines/tlbc-node` repository is empty. @schmir-at-bb are there already any efforts to adapt the _CircleCI_ flow and push to the according branches/tags?

weilbith:
So the _CircleCI_ configuration is already adopted to the new name. Therefore the release version must be just pushed.

PR #525, opened 2019-10-30
Use release docker images for tlbc quickstart

weilbith:
**Note:** the quickstart for the `tlbc` network will not work for the moment, since there is no `tlbc-node:release` image so far. This must be provided in future when the setup is ready.

Closes: #464

### DIFF
--- a/tools/quickstart/quickstart/configs/tlbc/docker-compose.yaml
+++ b/tools/quickstart/quickstart/configs/tlbc/docker-compose.yaml
@@ -6,7 +6,7 @@
 version: "2.4"
 services:
   home-node:
-    image: trustlines/tlbc:release
+    image: trustlines/tlbc-node:release
     restart: always
     stop_grace_period: 3m
     labels:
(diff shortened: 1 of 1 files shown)


tools/quickstart/quickstart/configs/tlbc/docker-compose.yaml
```diff
@@ -6,7 +6,7 @@
 version: "2.4"
 services:
   home-node:
-    image: trustlines/tlbc:release
+    image: trustlines/tlbc-node:release
```
berndbohmeier: This tag does not exist yet, but we need it for testing, so I think we should here just use :master for now.

weilbith: But this tag [doesn't exit either](https://hub.docker.com/r/trustlines/tlbc-node/tags). Or what do I miss? I though for the tests we still stick to laika. And tlbc will then contain already the validators of the auction. :thinking: 
Sry for the circumstances...

berndbohmeier: The tag exists on tlbc-node-next

PR #540, opened 2019-11-04
Check address in parity wrapper

berndbohmeier:
Check the address in the parity wrapper with a simple regex.
This would have cought an
error we had when the address was actually empty.


schmir-at-bb:
merged in 21ab0bf. I failed to include the original commit, because pre-commit forced to me to switch branches (it didn't let me commit to master).

Issue #462, opened 2019-10-14
Write Update process documentation


berndbohmeier:

berndbohmeier:
Also document the steps like 
- one a PR to branch x
...
- get an approval by ..

jannikluhn:
Started writing down steps for the release process:

```
# Release process

0) All changes have been merged into master
1) Create and merge a PR from master to pre-release.
2) Review changes as it is appropriate (at least the chain spec file). Other changes may be ignored depending on the changes (note that everything should have been reviewed already when it was merged into master)
3) Merge the PR. CircleCI will build new docker images and push them to docker hub under the name "trustlines/tlbc-node-next:pre-release" (or "trustlines/tlbc-testnet-next:pre-release" for testnet releases)
2) Test the docker image appropriately, at least:
   - Check that the chain spec file in the docker container is the right one
   - Start a new node and check that it connects to the network and starts syncing the chain
   - Take an existing setup started with the quickstart script (fully synced and connected to the network) and replace the running node with the new image. Check that it still connects to the network and stays at the top of the chain.
3) Make a PR from pre-release to release
4) Ensure that the testing steps have been performed, that no additional changes have been made, and merge the PR.
5) Check that the image is pushed to docker hub under "trustlines/tlbc-node:release" (or "trustlines/tlbc-testnet-node:release") and that a quickstart setup downloads it. Check that an updated node finds peers and continues processing the chain. Check via the netstats page that the network is still healthy.
```

schmir-at-bb:
I've just opened https://github.com/trustlines-protocol/blockchain/pull/511, which describes the release process for tools/bridge.
We should have a short talk about how we handle the setuptools_scm based projects.

schmir-at-bb:
https://github.com/trustlines-protocol/blockchain/pull/510 should also be taken into account. We'd like to maintain a version for the docker images. And that PR implements version handling.

andfletcher:
@schmir-at-bb to write down monitoring documentation

jannikluhn:
bridge release documentation is done: https://github.com/trustlines-protocol/blockchain/blob/master/tools/bridge/README-dev.md#release-checklist

jannikluhn:
Missing #581 and #582

schmir-at-bb:
done

PR #542, opened 2019-11-05
Write Bridge Release Checklist

jannikluhn:
Does not close but contributes to closing #462

### DIFF
--- a/tools/bridge/README-dev.md
+++ b/tools/bridge/README-dev.md
@@ -57,3 +57,26 @@ and open a PR for that branch.
 
 After a release has been made, we should upgrade the version on the
 master branch with `bumpversion patch`.
+
+## Release Checklist
+
+0. Prerequsisite: The master branch contains the version to be
+   released
+1. Open and merge a PR master -> pre-release (no review needed). This
+   will built the docker image and push it to Docker Hub under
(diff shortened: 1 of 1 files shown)


tools/bridge/README-dev.md
```diff
@@ -57,3 +57,29 @@ and open a PR for that branch.
...
+## Release Checklist
+
+0. Prerequsisite: The master branch contains the version to be
+   released
+1. Open and merge a PR master -> pre-release (no review needed). This
+   will built the docker image and push it to Docker Hub under
+   `trustlines/bridge-next:pre-release`.
+2. Test the pre-release:
+   - Pull the newly built image from Docker Hub.
+   - Check that the bridge config file looks as expected.
```
berndbohmeier: What bridge-config file?

jannikluhn: Ah, your right, I was thinking about these https://github.com/trustlines-protocol/blockchain/blob/master/tools/quickstart/quickstart/configs/laika/bridge-config.toml

But apparently we don't distribute them as part of the bridge, just in the quickstart script. I'll remove this, as well as the last sentence about testnets (which doesn't make sense if we don't have anything testnet related in there).

Issue #532, opened 2019-11-04
Ask if user want to set different author address


berndbohmeier:
Ask the user: "Do you want to receive mining rewards to a different address you own? This increases security of your funds. For more information see: http://gethelp.dsfsfsf"

Please insert an address where you have full control over on the >Trustlines Blockchain/ Laika testnet<. 

If he answers yes, ask for the address and set up parity so that it sends mining rewards to that address.

PR #543, opened 2019-11-05
Author Address

jannikluhn:
Closes https://github.com/trustlines-protocol/blockchain/issues/532

It doesn't work right now. I get the error "Unkown argument: --author" from parity. I'm not sure what's wrong: I checked that the environment variables are configured properly and from there on I just do the same for "author" as is already done for "validator". "author" is the right name according to the documentation and the changelog doesn't mention it, so it shouldn't have been renamed.

Could this the same issue you encountered, @berndbohmeier ?

### DIFF
--- a/tools/quickstart/quickstart/cli.py
+++ b/tools/quickstart/quickstart/cli.py
@@ -239,6 +239,7 @@ def run(
         )
     )
     validator_account.setup_interactively(base_dir=base_dir, chain_dir=chain_dir)
+    validator_account.setup_author_address(setup_name=setup_name, base_dir=base_dir)
     monitor.setup_interactively(base_dir=base_dir)
     bridge.setup_interactively(
         base_dir=base_dir,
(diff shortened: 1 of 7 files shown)


schmir-at-bb:
same error, yes. fix has been merged to master in c7b11cc.

jannikluhn:
The error persists even after rebasing, so it seems to be a separate issue. Investigating...

jannikluhn:
This seems to work now, but requires both the blockchain image to be rebuilt (as the wrapper script was updated).

schmir-at-bb:
So, by default we burn the block rewards on the zero address?

jannikluhn:
No, by default it goes to the validator address (https://github.com/trustlines-protocol/blockchain/pull/543/files#diff-9aa8e06f1aece82472dc89a029529a57R161). The zero address is only used for non-validators so that the parity wrapper has a valid argument to parse.

berndbohmeier:
Working on it

PR #548, opened 2019-11-06
Adapt the quickstart Laika bridge config for TTT

weilbith:
**Warning:** The home bridge is not funded at the moment.

Closes: #538
See also: trustlines-network/bridge-testing#4


weilbith:
Blocked until the verification problem of the `HomeBridge` contract has been solved. It will probably end up with a new address.

PR #550, opened 2019-11-07
Write release process for blockchain images

jannikluhn:
Very similar to the [one for the bridge](https://github.com/trustlines-protocol/blockchain/blob/master/tools/bridge/README-dev.md#release-checklist).

I've also clarified/fixed the branch names in the bridge release process description (add `bridge/` prefix) to distinguish from the ones for the blockchain.

### DIFF
--- a/chain/README-dev.md
+++ b/chain/README-dev.md
@@ -1,0 +1,32 @@
+# Developer Documentation
+
+## Release Checklist
+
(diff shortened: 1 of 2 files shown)


README-dev.md
```diff
@@ -0,0 +1,32 @@
+# Developer Documentation
```
berndbohmeier: should this not maybe be in chain/

Issue #556, opened 2019-11-07
Handle in quickstart that a key exists with different name than account.json


berndbohmeier:
Looks like an old Laika setup will have a key named `UTC-...`. We need to handle that case and move it to `account.json`.

berndbohmeier:
Done

PR #560, opened 2019-11-08
WIP: Handle legacy validator key files of format UTC-*

cducrest:
Handle legacy validator key files of format UTC-* by renaming them "account.json"

related: https://github.com/trustlines-protocol/blockchain/issues/556

If I run the quickstart with an old master commit (I randomly took https://github.com/trustlines-protocol/blockchain/commit/754cd7918d9e19c70154690e18e0728ce4c63cc0), and then run it with this branch, I get a permission denied error as quickstart cannot rename the legacy key file. Have to find another way.

### DIFF
--- a/tools/quickstart/quickstart/cli.py
+++ b/tools/quickstart/quickstart/cli.py
@@ -237,6 +237,7 @@ def run(
                 ),
             )
         )
+        + "\n"
     )
     validator_account.setup_interactively(base_dir=base_dir, chain_dir=chain_dir)
     validator_account.setup_author_address(setup_name=setup_name, base_dir=base_dir)
--- a/tools/quickstart/quickstart/cli.py
+++ b/tools/quickstart/quickstart/cli.py
@@ -235,9 +235,9 @@ def run(
                     "restarted and no configuration will be overwritten. It is possible to enable "
                     "additional components that you have chosen not to configure in earlier runs."
                 ),
+                "",
             )
         )
-        + "\n"
     )
     validator_account.setup_interactively(base_dir=base_dir, chain_dir=chain_dir)
     validator_account.setup_author_address(setup_name=setup_name, base_dir=base_dir)
--- a/tools/quickstart/quickstart/validator_account.py
+++ b/tools/quickstart/quickstart/validator_account.py
@@ -178,24 +178,23 @@ def get_author_address(base_dir) -> str:
 
 
 def handle_legacy_validator_key_file(base_dir, chain_dir):
-    legacy_key_file_paths = legacy_validator_key_paths(base_dir, chain_dir)
-    key_file_paths = legacy_key_file_paths.copy()
     if validator_key_file_exists(base_dir, chain_dir):
-        key_file_paths.append(validator_key_file_path(base_dir, chain_dir))
+        return
 
-    if len(key_file_paths) > 1:
+    legacy_key_file_paths = legacy_validator_key_paths(base_dir, chain_dir)
+    if len(legacy_key_file_paths) > 1:
         error_message = fill(
-            f"There are multiple files that correspond to validator keys: {key_file_paths}. "
-            f"Please make sure there is at most one key file in {os.path.join(base_dir, KEY_DIR, chain_dir)} "
-            f"and restart the quickstart."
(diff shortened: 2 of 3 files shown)


tools/quickstart/quickstart/cli.py
```diff
@@ -237,6 +237,7 @@ def run(
                 ),
             )
         )
+        + "\n"
```
schmir-at-bb: Please add an empty string at the of the "\n".join call instead.

tools/quickstart/quickstart/validator_account.py
```diff
@@ -171,3 +175,38 @@ def get_author_address(base_dir) -> str:
             return author_address_file.read()
     except FileNotFoundError:
         return get_validator_address(base_dir)
+
+
+def handle_legacy_validator_key_file(base_dir, chain_dir):
+    legacy_key_file_paths = legacy_validator_key_paths(base_dir, chain_dir)
+    key_file_paths = legacy_key_file_paths.copy()
+    if validator_key_file_exists(base_dir, chain_dir):
```
schmir-at-bb: Please do this check at the beginning of the function and if the validator key file exists just return.
That way you also don't have to copy the legacy_key_file_paths.

tools/quickstart/quickstart/validator_account.py
```diff
@@ -171,3 +175,38 @@ def get_author_address(base_dir) -> str:
             return author_address_file.read()
     except FileNotFoundError:
         return get_validator_address(base_dir)
+
+
+def handle_legacy_validator_key_file(base_dir, chain_dir):
```
berndbohmeier: The check and message is to strinct. It is perfectly fine to have more then one keystore file, just one has to be named account.json

so the check should be:
`if not validator_key_file_exist and legacy_file exist:
     if key_file_paths > 1:
           print("bla bla can not handle upgrading from one than more keyfiles, please rename one to account.json')
     else:
         rename
```

File: tools/quickstart/quickstart/validator_account.py (hunk 2 above)
Comment (cducrest) [reply tools/quickstart/quickstart/validator_account.py]: I can do that, but then if there is a validator key file in the format `account.json` and one in the format `UTC-*` then we will arbitrarily take the `account.json` one into account and ignore the other one. The idea there was to raise an error if this occurs.
Please confirm you still want me to change it and ignore the case where there are two keys.

Comment (cducrest) [reply tools/quickstart/quickstart/validator_account.py]: I guess that's also what Bernd wants, so I'll do that.

Comment (schmir-at-bb) [reply tools/quickstart/quickstart/validator_account.py]: That is fine, at least I fail to see a problem and Bernd has requested the same below. So, yes, please do it.

File: tools/quickstart/quickstart/validator_account.py (hunk 3 above)
Comment (cducrest) [reply tools/quickstart/quickstart/validator_account.py]: So you'd say it is fine if there are two keys in this directory and we just ignore arbitrarily one? Wouldn't that be confusing to the user?

Comment (berndbohmeier) [reply tools/quickstart/quickstart/validator_account.py]: I dont know what you mean by we ignore the other one. We need an account.json file, because that is the one that will be used by the bridge. So if there is account.json file we are already good.

File: tools/quickstart/quickstart/validator_account.py [hunk 4]
```diff
@@ -171,3 +175,44 @@ def get_author_address(base_dir) -> str:
...
+    if validator_key_file_exists(base_dir, chain_dir):
+        return
+
+    legacy_key_file_paths = legacy_validator_key_paths(base_dir, chain_dir)
+    if len(legacy_key_file_paths) > 1:
+        error_message = fill(
+            f"There are multiple files that correspond to validator keys: {legacy_key_file_paths}. "
+            f"Please make sure there is at most one key file in {os.path.join(base_dir, KEY_DIR, chain_dir)} "
```
berndbohmeier: I would write
There are multiple files that correspond to validator keys: {legacy_key_file_paths}. "
            f"Please make sure there is one key file in {os.path.join(base_dir, KEY_DIR, chain_dir)} called 'account.json', so we know which one to select for the bridge. "

Because having multiple key files should be fine as I said.

Issue #559, opened 2019-11-08
Remove possibility of using block author


cducrest:
As we noticed that using parity with the `--author` argument to send block reward to a different address from the signer address does not work, we need to remove/disable this option from the quickstart script.

schmir-at-bb:
I'd say please try to keep the code, but disable it via a flag or something like that.

berndbohmeier:
Done

PR #561, opened 2019-11-11
Don't configure author address

jannikluhn:
Parity unfortunately ignores this argument

### DIFF
--- a/tools/quickstart/quickstart/cli.py
+++ b/tools/quickstart/quickstart/cli.py
@@ -239,7 +239,8 @@ def run(
         )
     )
     validator_account.setup_interactively(base_dir=base_dir, chain_dir=chain_dir)
-    validator_account.setup_author_address(setup_name=setup_name, base_dir=base_dir)
+    # not in use because of https://github.com/paritytech/parity-ethereum/issues/11246
+    # validator_account.setup_author_address(setup_name=setup_name, base_dir=base_dir)
     monitor.setup_interactively(base_dir=base_dir)
     bridge.setup_interactively(
         base_dir=base_dir,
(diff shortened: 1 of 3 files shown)


jannikluhn:
Closes #561

schmir-at-bb:
> I'd say please try to keep the code, but disable it via a flag or something like that.

Was there a good reason why you didn't follow my suggestion from #559 here or did you just not see it?

jannikluhn:
I only saw your suggestion after I already removed everything, but I also don't see why we should keep the code. I don't think Parity will implement the author option any time soon and even if they do we can still recover it from the git history. Why would we want to keep unused code?

schmir-at-bb:
> I don't think Parity will implement the author option any time soon and even if they do we can still recover it from the git history.  Why would we want to keep unused code?

I do think it would have been easier to toggle a feature flag than digging into the git history. 
All we would have to do is not call into validator_account.setup_author_address. My assumption would be that they eventually fix that issue.

berndbohmeier:
Please do what ralf suggested.

tools/quickstart/quickstart/validator_account.py
```diff
@@ -71,32 +68,6 @@ def make_required_dirs(base_dir, chain_dir):
     os.makedirs(os.path.join(base_dir, KEY_DIR, chain_dir), exist_ok=True)
 
 
-def setup_author_address(setup_name, base_dir):
```
berndbohmeier: Do not remove, but add a comment that this can not be used right now because of a bug in parity and a link to the bug report.

tools/quickstart/quickstart/cli.py
```diff
@@ -239,7 +239,6 @@ def run(
         )
     )
     validator_account.setup_interactively(base_dir=base_dir, chain_dir=chain_dir)
-    validator_account.setup_author_address(setup_name=setup_name, base_dir=base_dir)
```
berndbohmeier: Just comment it out or use a flag with a hint why this does not work and a link to the bug report.

jannikluhn:
Alright, restored the code and commented out the unused parts, but I'd bet all my testnet TLCs that this won't ever be needed.

berndbohmeier:
Maybe not in this form, but if parts of the code count you will lose.

Issue #568, opened 2019-11-11
Change gas limit in chains spec and config


cducrest:
Change values to:
`minGasLimit` to 8_000_000
`gasLimit` of genesis to 8_000_000
keep `gasLimitBoundDivisor` as it is
explicitely set target gas limit of validators to 8_000_000 via `--gas-floor-target=8000000` argument or `
gas_floor_target = "8000000"` under `mining` in the config file.

PR #570, opened 2019-11-12
Adopt tlbc gas settings and behavior

weilbith:
Closes: #568


weilbith:
Just as note: I tested it locally and the gas limit for the genesis is enough for the pre-compiled contracts (just to be sure).

PR #583, opened 2019-11-13
Update README-dev.md

berndbohmeier:
I tried to fix the flow docu, is that what we want?
This will always result in merge conflicts. 
If we would merge back `release` , we would not end with merge conflicts I think. 
What was the reason again to not merge back?

### DIFF
--- a/tools/bridge/README-dev.md
+++ b/tools/bridge/README-dev.md
@@ -35,7 +35,7 @@ git checkout bridge/pre-release
 git pull
 git merge -m 'Merge with master branch' origin/master
 cd tools/bridge
-bumpversion build
+bumpversion release
 ```
 and open a PR for that branch.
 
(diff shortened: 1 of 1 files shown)


jannikluhn:
I guess if we'd merge back to master we'd end up with production release numbers in master. Not sure if that's a big deal, but probably not what we want. If we don't care about this, we could just bump the version once in master and then do the normal `->pre-release->release`, no?

berndbohmeier:
> I guess if we'd merge back to master we'd end up with production release numbers in master.

I dont know what you mean by that, master is only pointer to a single commit. 
If you mean by that that master references to some commit with a release version number, I dont see the issue, because that does not only have the version number, but actually is the release commit. 

> If we don't care about this, we could just bump the version once in master and then do the normal -> pre-release->release, no?

I also dont understand this, is that pro merging to master and done additionally or just another solution? Bumping from which version to which version?

jannikluhn:
>  If you mean by that that master references to some commit with a release version number, I dont see the issue, because that does not only have the version number, but actually is the release commit.

Yes, but (presumably) we would not immediately update this version number in the next commit. So later commits would have still have the same version number, even though there are changes. This is unavoidable, but we could restrict this to dev version numbers. As I said though I'm not sure if this is problematic (after all it's just a number, it's not actually released).

> I also dont understand this, is that pro merging to master and done additionally or just another solution? Bumping from which version to which version?

1) bump in master from the current to the next (release) version
2) merge ->pre-release->release
3) As there are no changes in the `pre-release` and `release` branches, they point to the same commit as master, so merging back to master is unnecessary.

berndbohmeier:
> Yes, but (presumably) we would not immediately update this version number in the next commit.

Thats what I would do, actually before merging to master, so that master always has dev version numbers. 

> bump in master from the current to the next (release) version ...

Then we dont need a version number if the version number does not represent anything. 
The purpose is to know what you are running, so the version number should at least tell you if you run a dev version or a proper version.

berndbohmeier:
I think for further discussion we should have a call.

schmir-at-bb:
After the discussion yesterday I've gone ahead and merged back the bridge/release branch to master in the blockchain repository. We need to do this for the projects that use setuptools_scm, so it makes sense to do this for all projects.

PR #585, opened 2019-11-13
Reduce foreign node requests to work better with hosted nodes

### LINKED ISSUE #576: Investigate number of bridge requests
Infura reports that I do 150k requests per day. I assume this is the bridge client. 
I thought we only to a request every 5 seconds, so we should investigate this.

berndbohmeier:
Closes #576 

This should reduce the foreign node requests from daily ~150k to ~40k. 

This comes with the trade of of losing one safty check, which I think is not necessary, and
a doubled poll interval of now 10s. Because 10s is still under the 15s block time of Görli and Mainnet, this should not be a problem.

### DIFF
--- a/tools/bridge/bridge/event_fetcher.py
+++ b/tools/bridge/bridge/event_fetcher.py
@@ -11,6 +11,8 @@ from bridge import node_status
 from bridge.events import ChainRole, FetcherReachedHeadEvent
 from bridge.utils import sort_events
 
+NODE_STATUS_CACHE_TIME_SECONDS = 1
+
 
 class EventFetcher:
     def __init__(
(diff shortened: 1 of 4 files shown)


tools/bridge/bridge/event_fetcher.py
```diff
@@ -54,14 +59,24 @@ def __init__(
...
+        if (
+            node_status is None
+            or time.time() - self._node_status_timestamp
+            > NODE_STATUS_CACHE_TIME_SECONDS
+        ):
+            # To reduce number of node calls we cache the result
+            self._node_status = self._retrying.call(
+                node_status.get_node_status, self.web3
+            )
+            self._node_status_timestamp = time.time()
```
schmir-at-bb: I would probably make the timestamp a field in `node_status.NodeStatus`. But we can also merge as is.

PR #588, opened 2019-11-14
Bridge/prepare/v1.0.0

berndbohmeier:


schmir-at-bb:
This should be merged to the bridge/pre-release branch, right?

berndbohmeier:
yes, ofc.

PR #606, opened 2020-01-02
Switch to setup.cfg src based layout

schmir-at-bb:


tools/bridge-deploy/Makefile
```diff
@@ -37,7 +37,7 @@ $(VIRTUAL_ENV):
 	python3 -m venv $@
 
 clean:
-	rm -rf build .tox .mypy_cache .pytest_cache */__pycache__ *.egg-info
-	rm -f .installed .compiled
+	rm -rf build .tox .mypy_cache .pytest_cache */__pycache__ */*/__pycache__ *.egg-info */*.egg-info
```
weilbith: I'm just curious: Why only one or two levels and not unlimited? :thinking: 
(the same in the `Makefile`s below)

```suggestion
	rm -rf build .tox .mypy_cache .pytest_cache **/__pycache__ **/*.egg-info
```

schmir-at-bb: I think that's only valid in zsh.

weilbith: Nope. That would not work for _git_ I guess. A snippet from `man gitignore`:

```
Two consecutive asterisks ("**") in patterns matched against full pathname may have special meaning:
 - A leading "**" followed by a slash means match in all directories. For example, "**/foo" matches file or directory "foo" anywhere, the same as pattern "foo". "**/foo/bar" matches file or directory "bar" anywhere that is directly under directory "foo".
 - ...
```

schmir-at-bb: why do you cite the gitignore manpage?

I'm using zsh for the first command here:

```
~/trustlines/protocol/blockchain (git)-[master] % echo **/__init__.py
contracts/tests/bridge/__init__.py contracts/tests/__init__.py tools/auction-deploy/src/auction_deploy/__init__.py tools/auction-deploy/tests/__init__.py tools/bridge-deploy/src/bridge_deploy/__init__.py tools/bridge-deploy/tests/__init__.py tools/bridge/src/bridge/__init__.py tools/bridge/tests/__init__.py tools/quickstart/src/quickstart/__init__.py tools/quickstart/tests/__init__.py tools/validator-set-deploy/src/validator_set_deploy/__init__.py tools/validator-set-deploy/tests/__init__.py
~/trustlines/protocol/blockchain (git)-[master] % bash
bbralf@zwackelmann:~/trustlines/protocol/blockchain$ echo **/*.py
**/*.py
```

weilbith: Because I'm stupid and where not fully focused on that but had other things in mind which were gitignore related. I'm sry. :sweat:

Issue #611, opened 2020-03-02
Update Laika chain specification file for new fork


weilbith:
- deploy new validator contract and add the addresses by trustlines-network/project#820
- reference contract address in new entry of the chain spec
- find a proper block number when to fork
- exchange bootstrap nodes in favor

weilbith:
The old bootstrap nodes will not be needed anymore, since they new validator nodes can act as such. I extended the requirements for that.

weilbith:
The new validator contract has been deployed [here](https://explore.laika.trustlines.foundation/address/0x7a393ffc122a61f8249c5571ad8992038a0f662f/read_contract). It is verified and the to read validator list should match the following of #820:

  - `0x116c0fe479f206b8e44c0ac519886484226e2eda`
  - `0x9523cce7eb06770af7fa889db23bfff1385db686`
  - `0xe5b8ea4551097c79fb84544d6b05b703301de9c7`
  - `0xe1dda141bf193b6faaf203923cf3520e1fbbdfe5`
  - `0x3b447ecc92c8d462ec9bca48c9ba1157996ff6a5`

weilbith:
This is a small script to calculate the block number for a fork to a specific date. The variables must be changed for the current situation. Only the parameter of the current block number must not be set. This script may be to excessive, but the intention was to keep it easy to verify.

The calculated value wobbles depending on the time of execution by 1 or two blocks in worst case. The average is `5686249`. So I would maybe suggest to use one block higher to have to have it "guaranteed" not before the intended fork date.


```sh
#! /bin/bash

# Parameters to set
validator_count_overall=23
validator_count_active=6
step_duration_seconds=5
date_fork='2020-03-31 00:00:00' # UTC time zone

# Extract current block number of RPC request and convert hex to int.
block_number_response_raw="$(curl --silent --data '{"method":"eth_blockNumber","params":[],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST https://laika.rpc.anyblock.tools:443)"
block_number_head_hex="$(echo "$block_number_response_raw" | awk -F '"' '{print $8}')"
block_number_head_hex="${block_number_head_hex#0x*}" # Remove leading 0x
block_number_head_hex="${block_number_head_hex^^}"   # Letters to upper case
block_number_head="$(echo "ibase=16; $block_number_head_hex" | bc)"

date_fork_seconds=$(TC=utc date +'%s' -d "$date_fork")
[...]

PR #612, opened 2020-03-09
Adjust Laika chain spec for validator fork

weilbith:
Just as recap: this PR includes two forks with an offset of 100 blocks. The first of them is intended to happen at the `2020-03-24 11:59:00 ECT`. I used [this script](https://github.com/trustlines-protocol/blockchain/issues/611#issuecomment-600004644) to calculate the guessed block number according to the network performance during the last 200.000 blocks.
The first fork is to a static JSON array, because such does not need finality in contrast to a contract. During the 100 blocks with the new validator, finality will be restored and allows to switch to the validator set contract. This contract is verified at the block explorer. Both address lists are equal. The bootnodes link to the new 5 validators. I tested it with all of them, adding them manually via JSON RPC.

Closes: #611


weilbith:
Note: valiator #5 is not available at the moment due to some DO issues. I'm
about to fix those soon.
Checkout the netstats page for their status.

weilbith:
Node is up and running again. :+1:

PR #618, opened 2020-05-07
[WIP] Add TokenValidatorAuction where one can bid with token

cducrest:
No other functionality added, cannot withdraw bids, lock, etc ...


contracts/contracts/auction/TokenValidatorAuction.sol
```diff
@@ -0,0 +1,71 @@
...
+        uint slotPrice = currentPrice();
+        require(
+            auctionnedToken.allowance(msg.sender, address(this)) >= slotPrice,
+            "The sender did not allow for the auction to withdraw enough tokens"
+        );
+        require(
+            auctionnedToken.balanceOf(msg.sender) >= slotPrice,
+            "The sender does not have enough tokens to bid"
+        );
+        auctionnedToken.transferFrom(msg.sender, address(this), slotPrice);
```
berndbohmeier, line 56: This is in theory open for reentrency attacks. This is not really the case for us, as we cant trust our token, but I would still prefer that you follow the recommended order of actions (https://consensys.github.io/smart-contract-best-practices/known_attacks/#reentrancy-on-a-single-function)

contracts/contracts/token/AuctionnableTokenInterface.sol
```diff
@@ -0,0 +1,76 @@
+pragma solidity ^0.5.8;
+
+contract AuctionnableTokenInterface {
```
berndbohmeier, line 3: Why can that not just be a erc20 token interface?

cducrest: Because I felt it was more work to find a clean 0.5.8 erc20 token interface.

berndbohmeier: I dont understand. The only thing that does not seam to be erc20 token is the name. 
Can we not rename it to `ERC20TokenInterface` or `TokenInterface`
I just dont know what a` AuctionnableTokenInterface` should be.

cducrest: Isn't there some missing functions in the interface to comply with the erc20 token?

contracts/contracts/auction/TokenValidatorAuction.sol
```diff
@@ -0,0 +1,49 @@
+pragma solidity ^0.5.8;
+
+import "../lib/Ownable.sol";
+import "./TokenDepositLocker.sol";
+import "../token/IERC20.sol";
+import "./BaseValidatorAuction.sol";
+
+
+contract TokenValidatorAuction is BaseValidatorAuction {
+    IERC20 public auctionnedToken;
```
berndbohmeier, line 10: Can we use simply `token` or `bidToken` or so?
auctionnedToken sounds like we are auctioning the token to me.

contracts/tests/bridge/test_measure_gas_home_bridge.py
```diff
@@ -93,6 +95,7 @@ def __call__(self, n, fail_ok=False):
     return ConfirmNth()
 
 
+@pytest.mark.skip
```
berndbohmeier, line 98: Can you please add a reason right here why we skip them?
And also add a `# TODO`

berndbohmeier: see https://docs.pytest.org/en/latest/skipping.html#skipping-test-functions

berndbohmeier:
Overall I think it is a bid too much code. The tests seems fine, but there seems to be a lot of fixture code that I believe could be condensate somehow.  The more code the more work to maintain it. 
But we can have a look later.

contracts/contracts/auction/TokenValidatorAuction.sol
cducrest: bidToken sounds good

PR #622, opened 2020-05-12
Token deposit locker

### LINKED ISSUE #619: Deposit locker can lock TLN
Deposit locker can receive TLN from the auction contract and lock them.
Users can withdraw TLN from deposit locker after lock period.

berndbohmeier:
closes #619

### DIFF
--- a/contracts/contracts/auction/TokenDepositLocker.sol
+++ b/contracts/contracts/auction/TokenDepositLocker.sol
@@ -1,0 +1,36 @@
+pragma solidity ^0.5.8;
+
+import "./DepositLocker.sol";
+import "../token/TrustlinesNetworkToken.sol";
+
+/*
+  The ETHDepositLocker contract locks ETH deposits
+
+  For more information see DepositLocker.sol
+*/
+
+contract TokenDepositLocker is DepositLocker {
+    TrustlinesNetworkToken public token;
+
+    constructor(TrustlinesNetworkToken _trustlinesNetworkToken) public {
+        require(
+            address(_trustlinesNetworkToken) != address(0),
+            "Token contract can not be on the zero address!"
+        );
+        token = _trustlinesNetworkToken;
+    }
+
+    function _receive(uint amount) internal {
+        require(msg.value == 0, "Token locker contract does not accept ETH");
+        // to receive erc20 tokens, we have to pull them
+        token.transferFrom(msg.sender, address(this), amount);
+    }
+
+    function _transfer(address payable recipient, uint amount) internal {
+        token.transfer(recipient, amount);
+    }
+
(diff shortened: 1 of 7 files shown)


contracts/contracts/auction/TokenDepositLocker.sol
```diff
@@ -0,0 +1,36 @@
+pragma solidity ^0.5.8;
+
+import "./DepositLocker.sol";
+import "../token/TrustlinesNetworkToken.sol";
+
+/*
+  The ETHDepositLocker contract locks ETH deposits
```
cducrest, line 7: The TokenDepositLocker contract locks Token deposits

contracts/contracts/auction/TokenDepositLocker.sol
```diff
@@ -0,0 +1,36 @@
...
+
+  For more information see DepositLocker.sol
+*/
+
+contract TokenDepositLocker is DepositLocker {
+    TrustlinesNetworkToken public token;
```
cducrest, line 13: It should be better to use a token interface to keep it generic as it was the mentality we had on these contracts

contracts/contracts/auction/TokenDepositLocker.sol
```diff
@@ -0,0 +1,36 @@
...
+
+    constructor(TrustlinesNetworkToken _trustlinesNetworkToken) public {
```
cducrest, line 15: I think it is weird that the contract now has both a `constructor` and an `init`, I feel like it is less obvious to work with

contracts/tests/test_deposit_locker.py
```diff
@@ -220,11 +138,43 @@ def withdraw(self, depositor):
...
+            token_symbol,
+            token_decimal,
+            premint_token_address,
+            premint_token_value,
+        )
+
+        token_contract = deploy_contract(
+            "TrustlinesNetworkToken", constructor_args=constructor_args
+        )
+        constructor_args = (token_contract.address,)
```
cducrest, line 175: I would make the difference between deposit_constructor_args and token_constructor_args

contracts/contracts/auction/TokenDepositLocker.sol
berndbohmeier: Changed

contracts/contracts/auction/TokenDepositLocker.sol
berndbohmeier: Changed to only use init.

contracts/tests/test_deposit_locker.py
berndbohmeier: outdated now, as there is not deposit constructor anymore.

PR #624, opened 2020-05-19
Feature/token auction deploy

cducrest:


tools/auction-deploy/src/auction_deploy/core.py
```diff
@@ -1,12 +1,14 @@
-from typing import Dict, NamedTuple, Optional, Sequence
+from typing import Dict, NamedTuple, Optional, Sequence, Tuple
 
 from deploy_tools.deploy import (
     deploy_compiled_contract,
     increase_transaction_options_nonce,
     load_contracts_json,
     send_function_call_transaction,
 )
+from eth_tester.exceptions import TransactionFailed
```
berndbohmeier, line 9: core should not import stuff from eth tester

tools/auction-deploy/src/auction_deploy/core.py
```diff
@@ -152,6 +183,20 @@ def get_deployed_auction_contracts(
...
+def get_bid_token_address(web3, auction_address: str):
+    compiled_contracts = load_contracts_json(__name__)
+    auction_abi = compiled_contracts["TokenValidatorAuction"]["abi"]
+    auction = web3.eth.contract(address=auction_address, abi=auction_abi)
+    try:
+        return auction.functions.bidToken().call()
+    except BadFunctionCallOutput:
+        # Thrown by web3 when function does not exist on contract
+        return None
+    except TransactionFailed:
```
berndbohmeier, line 195: How did you test the BadFunctionCallOutput if this is raised by the tester?
I dont like that we here handle the tester specific, I would hope we can somehow remove this.

tools/auction-deploy/tests/conftest.py
```diff
@@ -1,17 +1,30 @@
...
+
+assert (
+    eth_tester.backends.pyevm.main.GENESIS_GAS_LIMIT < 8 * 10 ** 6
+)  # noqa: E402, E702 isort:skip
+eth_tester.backends.pyevm.main.GENESIS_GAS_LIMIT = (
+    8 * 10 ** 6
+)  # noqa: E402, E702 isort:skip
+
+
+import pytest  # noqa: E402, E702 isort:skip
```
berndbohmeier, line 17: it is fine that the stuff above have the ignore linting stuff, but that here every line has it is a bit much. Please find a way to only have to exclude the eth_tester lines.

tools/bridge/pytest
```diff
@@ -6,10 +6,10 @@ from gevent import monkey  # isort:skip
 
 monkey.patch_all()  # noqa: E402 isort:skip
 
-import re
-import sys
+import re  # noqa: E402, E702 isort:skip
```
berndbohmeier, line 9: same as above.

tools/auction-deploy/src/auction_deploy/core.py
cducrest: I tested the BadFunctionCallOutput by deploying an auction on goerli and calling status on it

tools/auction-deploy/tests/conftest.py
cducrest: Took me quite some time to figure it out. My understanding is that the flake8 ignore for `E402` has to be on every line that is an import below a non import statement or that the non import statement has to be on the same line as an import statement (for some reason).

Tell me if you like the current version better.

PR #630, opened 2020-06-16
Feature/upgrade parity

cducrest:
uses new builder for circleci not yet on master.

I think you can check commits individually, and if you don't like renaming `parity` to `openethereum` we can discard the last two.


chain/client_wrapper.sh
```diff
@@ -44,13 +44,13 @@
 #     - Needs the password file and the key-set. (see FILES)
 #
 # FILES
-#   The configuration folder for Parity takes place at /home/parity/.local/share/io.parity.ethereum.
+#   The configuration folder for used client takes place at /home/openethereum/.local/share/io.parity.ethereum.
```
berndbohmeier, line 47: the end `io.parity.ethereum` did not change?

berndbohmeier:
ok, we need to find out why the pre commit hooks fail, it is getting annoying.

chain/client_wrapper.sh
cducrest: That is what their documentation states, but it is lying. By testing, I see that it uses `io.parity.ethereum` if it exists, and `openethereum` otherwise.
https://openethereum.github.io/wiki/Configuring-Parity-Ethereum

berndbohmeier: okay, so thats sounds like it is only the fallback, so maybe we should switch to openethereum there too?

cducrest: Yes.

cducrest: I fixed it, the remaining places in the code where we have the word `parity` left are very few and easy to check, which should mean there is no mistake left. (config.toml under [parity], and the name of some rpc are still `parity_...`, the behaviour of which I checked locally with openethereum)

berndbohmeier: sounds good

chain/client_wrapper.sh
```diff
@@ -147,11 +147,11 @@ function parseArguments() {
       shift                # arg
       shift                # value
 
-    # Additional arguments for the Parity client.
-    # Use all remain arguments for parity.
-    elif [[ $arg == --parity-args ]] || [[ $arg == -p ]]; then
+    # Additional arguments for the client.
+    # Use all remain arguments.
+    elif [[ $arg == --client-args ]] || [[ $arg == -c ]]; then
```
berndbohmeier, line 152: I think we will need to handle `--parity-args` and `-p` for backwards compatibility

PR #644, opened 2020-07-18
Use upstream prettier pre-commit hook

schmir-at-bb:
That means you can stop maintaining the prettier fork with the integrated solidity plugin.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,12 +14,13 @@ repos:
     - id: shfmt
     - id: shellcheck
 
--   repo: https://github.com/trustlines-protocol/prettier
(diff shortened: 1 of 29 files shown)


cducrest:
How come we were needing it and no longer do?

schmir-at-bb:
> How come we were needing it and no longer do?

The first time I tried to use it I couldn't get prettier to work with the solidity plugin. So, I threw it together into this prettier fork.

 We did the same for shutter and I didn't want to rely on this fork. The trick is to pass the "--plugin=prettier-plugin-solidity" argument to make it work with the pre-commit managed node installation. I think I found this  in the  pre-commit issue tracker.

PR #649, opened 2020-09-04
Add option --expose-ports to quickstart to expose the home node ports

cducrest:
close: https://github.com/trustlines-network/project/issues/881

It is a slight pain to test.

The quickstart script does not work on master, since it pulls the latest released docker image for the home node, and that image does not use Openethereum, while the docker-compose file expects the version with Openethereum. This is to keep the repository coherent over functional.

To test this, you thus need to build the home node image you want to test (laika or tlbc) on master and tagging it the way the docker-compose file expects it. e.g. `cd chain && docker build -f tlbc/Dockerfile -t trustlines/tlbc-node:release .`.

You then need to prevent the quickstart script to pull the image and replace it by commenting lines 233-239 of `quickstart/src/quickstart/docker.py`.

After that you are free to test whatever you want, for example `quickstart tlbc --expose-ports` and test that the home node is reachable with `curl --data '{"method":"eth_blockNumber","params":[],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545`.

### DIFF
--- a/quickstart/src/quickstart/cli.py
+++ b/quickstart/src/quickstart/cli.py
@@ -77,6 +77,33 @@ def base_dir_option(**kwargs):
     )
 
 
+def confirm_expose_ports(ctx, param, value):
+    if value:
+        click.secho(
+            "Running the home node with exposed ports incurs risks.\n"
+            "If the machine on which the node is ran exposes the\n"
+            "ports itself the node will be queryable from outside.\n"
+            "With the wrong settings, this means funds could be stolen.\n\n"
+            "Only use this option if you know what you are doing.\n",
+            "You can later change the exposed ports via the docker-compose.override.yaml file",
+            fg="red",
+        )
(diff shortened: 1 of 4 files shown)


dakingha69 (approved):
Codewise looks good, functionality-wise also 👍 Tested it as you described. One thing we could think of is to add some information on how to expose custom ports on the host machine. I tested custom ports by changing the port mappings in `docker-compose.override.yaml`. Either just mention some words in the quickstart to the user on this or even add the possibility to set the ports via the CLI flag. Just a thought

cducrest:
I initially had it that the user could set up a port via flag but it's a pain to make it work with docker, because then I need to use environment variables in the docker-override.yml. If the user for example gives invalid inputs (ports already taken, same ports twice, ports lower than 80, no ports for WS, etc ...) the container will not start/the docker-compose file will be invalid and it is hard to debug for the user.

I'll had some docs

PR #650, opened 2020-09-04
Update pre-commit so that it runs mypy properly on individual packages

cducrest:
Pre-commit and circleci run mypy on all .py files within the repo.


dakingha69 (approved):
LGTM

